Issue #36, opened 2021-01-12
Speed up transitions after raster, standard, vector data collections


dalekreitler-bnl:
# Overview
* currently takes ~25-30 seconds after last image acquired to control returned
* reduce or eliminate sleeps
* parallelize independent steps that are currently performed in series

# Raster transition flow

![end_of_raster_flow 30Dec2020](https://user-images.githubusercontent.com/61419915/104248998-93d45500-5438-11eb-8546-7d5459bacc60.png)

# Proposed changes (examples)
* Stop detector, reset zebra, move goniometer while spot finding threads are running, affects rasters (https://github.com/dalekreitler-bnl/lsdc/commit/cdd48968aae3325eb8be7309230db08f322a268f)
* Parallel camera toggle/gov transition, affects rasters/standard/vector (https://github.com/dalekreitler-bnl/lsdc/commit/8cfa85aeb66d8624de81c0e5797253bc47ac1499)
* Reduce sleep before ispyb entry, affects standards/vectors (https://github.com/dalekreitler-bnl/lsdc/commit/96a3baa73ac568247c20c3d5c16d935739ed331e)
![end_of_raster_flow proposed_changes](https://user-images.githubusercontent.com/61419915/104249051-ab134280-5438-11eb-9b24-247ff94d4066.png)

# Raster specific
* xrecRasterFlag PV tells GUI clients to draw heat map and take snapshot for ispyb https://github.com/NSLS-II/lsdc/blob/5badd72469e455fcfbce3744fb1a2fa692b00464/lsdcGui.py#L3013-L3039
* currently xrecRasterFlag PV changed after governor transition, this causes long delay in completion of raster and display of results
* current order

>1. governor transition initiated (DA -> SA or DA -> DI for automated)
>2. xrecRasterFlag PV changes

Next steps part of same function

>3. heat map drawn
>4. ispyb rasterResult entered (not sure on exact timing of jpeg image acquired, needs lamp in though!)
 
![rasters](https://user-images.githubusercontent.com/61419915/104249063-b23a5080-5438-11eb-958e-c47bf9d40f06.png)

* put completed raster request to xrecRasterFlag PV before gov transition initiated on delayed Timer thread step 2 becomes a separate thread that starts a few seconds after step 1 and occurs in parallel now, see video below, works well but this delay will require tuning

https://user-images.githubusercontent.com/61419915/104249367-499fa380-5439-11eb-9381-919ccb0c6eb7.mov

[...]

PR #46, opened 2021-02-10
Earlier gov transitions and gov status pv checks at end of rasters

dalekreitler-bnl:
### **Order of steps in daq_macros.snakeRasterNormal after data acquisition**  (#36)
 
new_rasters

* Eliminates ispyb images with black background with earlier governor transition
* Timing approximately 3.5-5 seconds depending on tuning of delays
* Timing delays tunable with database parameters

### DIFF
--- a/bin/dump_blconfigs.py
+++ b/bin/dump_blconfigs.py
@@ -1,0 +1,44 @@
+import db_lib
+from datetime import date
+import sys
+
+'''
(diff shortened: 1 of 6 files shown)


JunAishima (approved):
all good! daq_lib is fixed by 1df048c, daq_macros is fixed by dcd4043 which was in the original set of commits. approving.

PR #51, opened 2021-02-16
add unlatchGov() command for FloCos

pelirrojo725:
A simple command to make the governor active for the FloCos during their recovery procedure. This ensures uniform training and smoother recovery procedures. The changes were added to daq.py. The command is unlatchGov().

### DIFF
--- a/daq_lib.py
+++ b/daq_lib.py
@@ -66,6 +66,10 @@ def destroy_gui_message():
   beamline_support.pvPut(gui_popup_message_string_pv,"killMessage")
 
 
+def unlatchGov(): # Command needed for FloCos to recover robot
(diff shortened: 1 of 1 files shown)


JunAishima (commented):
I think the check on the PV will probably come back as false because of the EPICS update time (unless it was already active). I would suggest printing the value before setting, then allowing the value to go through the Governor for at least 0.3 sec before getting its value back (just put in a sleep).

pelirrojo725:
Can we talk? I don't understand what you are saying.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

JunAishima (approved):
Looks good! thanks for the work!

PR #101, opened 2021-07-21
Eiger ophyd conversion

JunAishima:
updating existing LSDC code to remove code now in mxtools, and to use mxtools.scans functions


daq_macros.py
```diff
@@ -26,7 +26,10 @@
 from collections import OrderedDict
 from scans import (zebra_daq_prep, setup_zebra_vector_scan,
                    setup_zebra_vector_scan_for_raster,
-                   setup_vector_program)
+                   setup_vector_program, setup_eiger_exposure,
+                   setup_eiger_triggers, setup_eiger_arming,
+                   setup_eiger_stop_acquire_and_wiat
```
RobertSchaffer1, line 31: wiat -> wait

daq_macros.py
```diff
@@ -3201,7 +3202,7 @@ def zebraVecDaqSetup(angle_start,imgWidth,exposurePeriodPerImage,numImages,fileP
   exposureTimePerImage =  exposurePeriodPerImage - detector_dead_time
   yield from zebra_daq_prep()
```
RobertSchaffer1, line 3203: Does this (also others?) need to be changed to zebra_daq_prep(zebra) as well?

daq_macros.py
```diff
@@ -26,7 +26,10 @@
 from collections import OrderedDict
 from scans import (zebra_daq_prep, setup_zebra_vector_scan,
```
RobertSchaffer1, line 27: Where is scans being imported from now that the old file is deleted?

start_bs.py
```diff
@@ -3,7 +3,9 @@
 from ophyd import *
 from ophyd.mca import (Mercury1, SoftDXPTrigger)
 from ophyd import Device, EpicsMotor, EpicsSignal, EpicsSignalRO
-from zebra import Zebra
+from mxtools.zebra import Zebra
```
RobertSchaffer1, line 6: I found this in the AMX collection - is that where we are importing it from here? If so, will this work sensibly if this file is run by FMX?

daq_macros.py
JunAishima: this should be coming from mxtools as with zebra below

start_bs.py
JunAishima: We run LSDC using a Conda environment that contains multiple dependencies, like how we are doing the FMX profile_collection. We will need to add mxtools to the new environment. Ideally, we would make LSDC pip installable, but that is a while away. Until then, we have to remember that mxtools needs to be added to the MX environment.

Also, having mxtools available as a Conda package will make it easier to integrate into future MX Conda environments.

daq_macros.py
JunAishima: yes :( looks like I missed at least one... will need to check over things

start_bs.py
JunAishima: Looks like I answered a slightly different question - I put mxtools into NSLS-II-AMX because it's MX-specific (so shouldn't be in NSLS-II), but the NSLS-II-MX is used for processing software. We should be able to use it despite it being in the "wrong" organization.

daq_macros.py
JunAishima: fixed in https://github.com/NSLS-II/lsdc/pull/101/commits/258e1befc4d7365a64d93d84bf193806ed91a7e0

daq_macros.py
JunAishima: fixed in https://github.com/NSLS-II/lsdc/pull/101/commits/258e1befc4d7365a64d93d84bf193806ed91a7e0

PR #102, opened 2021-07-29
Robot api

JunAishima:
an attempt to make a reasonably clean interface for two completely different robot systems into LSDC.

I've extracted the current robot_lib that is AMX/FMX-specific into embl_robot, and developed a new module for the robot on NYX (denso_robot), which is controlled through an Ophyd object.

sorry for the messy commits and lots of corrections - I can re-do the work in a cleaner way for future history, but feedback on whether the direction of this work is correct will help evaluate where to go next.


start_bs.py
```diff
@@ -105,11 +109,18 @@ def filter_camera_data(camera):
...
     mercury.read_attrs = ['mca.spectrum', 'mca.preset_live_time', 'mca.rois.roi0.count',
                                             'mca.rois.roi1.count', 'mca.rois.roi2.count', 'mca.rois.roi3.count']
     vdcm = VerticalDCM('XF:17IDA-OP:FMX{Mono:DCM', name='vdcm')
     zebra = Zebra('XF:17IDC-ES:FMX{Zeb:3}:', name='zebra')
     vector_program = VectorProgram('XF:17IDC-ES:FMX{Gon:1-Vec}', name='vector_program')
 
+    robot = EMBLRobot()
+elif beamline == "nyx":
+    denso_robot = Robot("XF:19IDC-ES{Rbt:1}", name="robot")
+    robot = DensoRobot(denso_robot)
```
mrakitin, line 124: These names are a bit confusing, it'd be better to distinguish between the ophyd classes/instances and the robot API classes/instances.

start_bs.py
```diff
@@ -105,11 +109,18 @@ def filter_camera_data(camera):
...
+else:
+    raise Exception("Unknown beamline")
```
mrakitin, line 126:
```suggestion
    raise Exception(f"Unknown beamline: {beamline}")
```

mrakitin (commented):
Looks good overall besides formatting, which we agreed will be applied in the future PR in bulk for all the code.

start_bs.py
JunAishima: completely agree - will work on updating the Robot class name in mxtools and then clean up the code downstream from there

JunAishima: commit https://github.com/NSLS-II/lsdc/pull/102/commits/a53cacbad93d418692ec663ac1e2e8dc27ce887f - in nyxtools, have renamed the Robot class to DensoOphydRobot, given more descriptive names to the objects, and added a note about the DensoRobot, which is the robot_lib API-compatible object. hope this meets with your satisfaction.

mrakitin (approved):

PR #104, opened 2021-08-11
2021 2

JunAishima:
changes from late in the last shutdown and during the last cycle, so these have mostly been tested on the beamlines already.

### DIFF
--- a/bin/MT_LoadCheck.py
+++ b/bin/MT_LoadCheck.py
@@ -1,0 +1,47 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Mon Aug 29 10:37:32 2016
+
+@author: kun
+"""
+
+import robot as rob
+import time
+
+CMD_TIMEOUT = 500000
+
+
+try:
+
+    cmds=['Initialize','Cooldown']
+    for scmd in cmds:
+        tskStat, sampleStat, exception = rob.runATask(scmd, CMD_TIMEOUT)
+        if (tskStat.lower()!="done"):
+            raise Exception(scmd+" "+tskStat+" with exception "+exception)
+
+    cmds=['MT_loadCheck']
+    for i in range(1,2):
+        print( "================================== Plate",i, "=================================")
+
+        for n in range(1,49):
+            rob.par.execute("nSample",n)
+            rob.parChk.execute("nDummy")
+            if n!=float(rob.parChk.execute("nSample")):
+                raise Exception("Failed to set robot viable nSample")
+
+#            if n in [17,33]:
+#		rob.runATask("Cooldown", CMD_TIMEOUT)
+
+            for scmd in cmds:
+                tskStat, sampleStat, exception = rob.runATask(scmd, CMD_TIMEOUT)
+                if (tskStat.lower()!="done" and not tskStat.startswith("Done")):
+                    raise Exception(scmd+" "+tskStat+" with exception "+exception)
+
+
+    #rob.runATask("TraceSample", CMD_TIMEOUT)
+
+finally:
+    rob.runATask("TraceSample", CMD_TIMEOUT)
+    rob.runATask("Finish", CMD_TIMEOUT)
+
+
--- a/bin/MT_LoadCheck.py
+++ b/bin/MT_LoadCheck.py
@@ -43,5 +43,3 @@ try:
 finally:
     rob.runATask("TraceSample", CMD_TIMEOUT)
     rob.runATask("Finish", CMD_TIMEOUT)
-
-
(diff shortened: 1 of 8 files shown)


bin/MT_LoadCheck.py
```diff
@@ -0,0 +1,47 @@
...
+            raise Exception(scmd+" "+tskStat+" with exception "+exception)
+
+    cmds=['MT_loadCheck']
+    for i in range(1,2):
+        print( "================================== Plate",i, "=================================")
+
+        for n in range(1,49):
+            rob.par.execute("nSample",n)
+            rob.parChk.execute("nDummy")
+            if n!=float(rob.parChk.execute("nSample")):
```
RobertSchaffer1, line 29: I'm confused - why are we comparing a float and an int?

JunAishima (commented):
Robert, I fixed the type issue in the MT_LoadCheck.py script with commit https://github.com/NSLS-II/lsdc/pull/104/commits/22dd77d7af6d15a3f31b47367f290f0927851086

bin/MT_LoadCheck.py
```diff
@@ -0,0 +1,47 @@
...
+            raise Exception(scmd+" "+tskStat+" with exception "+exception)
+
+    cmds=['MT_loadCheck']
+    for i in range(1,2):
+        print( "================================== Plate",i, "=================================")
+
+        for n in range(1,49):
+            rob.par.execute("nSample",n)
+            rob.parChk.execute("nDummy")
+            if n!=int(float(rob.parChk.execute("nSample")))
```
RobertSchaffer1, line 29: Better, but is it still possible that this will round down (ie int takes 1.9999994 ->1 instead of 2)?

epics_det.py
```diff
@@ -196,7 +196,7 @@ def det_waitArmed(): #not even sure what this means.
     start_time = time.time()
     while (get_det_pv("armed_state") == 0):
       if time.time() - start_time > timeout:
-        logger.info('aborting arm')
+        logger.error('aborting arm')
```
RobertSchaffer1, line 199: Maybe add "Timed out" or some such?

bin/MT_LoadCheck.py
JunAishima: we are getting back a rounded floating point value from the underlying database as a string - so "2.50" or "2.0". I agree that this handling isn't ideal, but it is probably the best we can do with a general mechanism of returning numbers and imprecise floating point number formats.

epics_det.py
JunAishima: agree with this - fixed with commit https://github.com/NSLS-II/lsdc/pull/104/commits/a98e6cfa78b50173f1b72839359844628d3090ec

lsdcGui.py
```diff
@@ -3123,10 +3137,11 @@ def processBeamSize(self,beamSizeFlag):
...
     def processEnergyChange(self,energyVal):
-      if (energyVal<9000):
-        self.beamsizeComboBox.setEnabled(False)
-      else:
-        self.beamsizeComboBox.setEnabled(True)
+      if daq_utils.beamline != "amx":
+        if (energyVal<9000):
+          self.beamsizeComboBox.setEnabled(False)
+        else:
+          self.beamsizeComboBox.setEnabled(True)
```
RobertSchaffer1, line 3144: Consider adding 

`print("It's over 9000!") `

JunAishima: this is a conservative change that preserves the previous behavior. I won't change it now, but will consult with FMX to see what feedback users should get when the beam size is no longer changeable.

RobertSchaffer1 (approved):
I think it looks alright, there's a few style things that we could sort out later but otherwise I didn't notice issues. That said, it's all complicated enough that I have no real idea what other problems might arise - best to merge and test, I think. Wait for the second review though.

bin/MT_LoadCheck.py
```diff
@@ -0,0 +1,47 @@
...
+CMD_TIMEOUT = 500000
+
+
+try:
+
+    cmds=['Initialize','Cooldown']
+    for scmd in cmds:
+        tskStat, sampleStat, exception = rob.runATask(scmd, CMD_TIMEOUT)
+        if (tskStat.lower()!="done"):
+            raise Exception(scmd+" "+tskStat+" with exception "+exception)
```
RobertSchaffer1, line 20: I think these are caught right now by the try statement - add an except clause sending them to errout?

bin/MT_LoadCheck.py
```diff
@@ -0,0 +1,47 @@
...
+                    raise Exception(scmd+" "+tskStat+" with exception "+exception)
+
+
+    #rob.runATask("TraceSample", CMD_TIMEOUT)
+
+finally:
+    rob.runATask("TraceSample", CMD_TIMEOUT)
+    rob.runATask("Finish", CMD_TIMEOUT)
+
+
```
RobertSchaffer1, line 47: Blank space at end of file

mskinner5278 (approved):

PR #125, opened 2021-09-24
add lsdcRemote-related scripts

JunAishima:
* use these instead of local copies that have gotten
   out of sync with current LSDC code and conda enviornment
   requirements

### DIFF
--- a/bin/lsdcRemote
+++ b/bin/lsdcRemote
@@ -1,0 +1,1 @@
+script -a -c lsdcRemote.cmd -f lsdcRemoteLog.txt
--- a/bin/lsdcRemote_amx.cmd
+++ b/bin/lsdcRemote_amx.cmd
@@ -1,0 +1,16 @@
+#!/bin/bash -l
+export PROJDIR=/GPFS/CENTRAL/xf17id2/skinnerProjectsBackup/
+export CONFIGDIR=${PROJDIR}bnlpx_config/
+export LSDCHOME=${PROJDIR}lsdc_amx
+export PYTHONPATH=".:${CONFIGDIR}:/usr/local/edna-mx/mxv1/src:/usr/local/edna-mx/kernel/src:${LSDCHOME}:${PROJDIR}/RobotControlMerge"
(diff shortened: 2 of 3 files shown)


RobertSchaffer1 (commented):
Those files are pretty close to identical, are you sure we want two files and not one with a variable set?

bin/lsdcRemote
```diff
@@ -0,0 +1 @@
+script -a -c lsdcRemote.cmd -f lsdcRemoteLog.txt
```
RobertSchaffer1, line 1: I'm not super familiar with the script command, but I *think* this will try to run the "lsdcRemote.cmd" file which is now missing (having been renamed to "lsdcRemote_amx.cmd" and "lsdcRemote_fmx.cmd").

JunAishima: The procedure for creating the symlinks will help explain this. Probably should be documented!!

On both AMX and FMX, the lsdcServer and lsdcRemote commands are symlinked from /usr/local/bin without any changes to filename.

For the scripts lsdcServer_amx/fmx.cmd and lsdcRemote_amx/fmx.cmd, these files are symlinked and renamed to lsdcServer.cmd and lsdcRemote.cmd, as appropriate for the beamline.

This allows the same general script to be used for both/all beamlines. Eventually this should be added to documentation, and implemented in ansible jobs for installing LSDC

JunAishima:
> Those files are pretty close to identical, are you sure we want two files and not one with a variable set?

I have been making these files separate to reduce the number of conditions which increase the complexity of the code somewhat.

With so many other places where branching makes the code extremely complicated and hard to follow, I am happy to keep this code this way.

RobertSchaffer1:
> > Those files are pretty close to identical, are you sure we want two files and not one with a variable set?
> 
> I have been making these files separate to reduce the number of conditions which increase the complexity of the code somewhat.
> 
> With so many other places where branching makes the code extremely complicated and hard to follow, I am happy to keep this code this way.

Okay, works for me.

RobertSchaffer1 (approved):

mskinner5278 (approved):

PR #126, opened 2021-10-07
Albula opening changes

RobertSchaffer1:
- Albula opens automatically
 - fastDP moved to the staff screen
 - +-1 degree moves added


JunAishima:
I removed some code I had put in there - not sure what it was even for. otherwise, looks good.

PR #130, opened 2021-11-02
Processing program rpm locations

JunAishima:
As programs for the AMX/FMX cluster are now being installed by RPMs and the binaries and libraries are in standard locations or accessible from the commandline with just their names without full path names, update LSDC to remove the use of explicit non-standard command paths.


JunAishima:
The changes have worked as expected, so after some final checks, I will open this up for review

RobertSchaffer1:
Has a chunk of this already been merged? I'm seeing only the most recent changes on my files changed tab, and I had been seeing no change at all prior to the most recent push.

JunAishima:
As Rob has pointed out, this was actually included in https://github.com/NSLS-II/lsdc/pull/143 , which I merged with cluster node changes in the GUI and server. I will need to cherry-pick the last commit onto an active branch, but otherwise can close.

PR #143, opened 2021-11-16
Cluster nodes rpm program locations

### LINKED ISSUE #142: LSDC code changes for using new uranus cluster
- hostnames for cluster nodes - for parsing which nodes to show in GUI, and what to set when given input
- program calls use standard locations or just <program name> without full path instead of /usr/local/crys-local
- neggia library location is standard instead of within /usr/local/crys-local for raster processing

JunAishima:
Set of code fixes required within LSDC to allow LSDC to run and process data on the new uranus-cpu??? cluster. As part of this work, RHEL 8 was used on the new computers and RPMs were used to install the processing software, which can now be run by just using the program name instead of a full, nonstandard path.

fixes #142


daq_macros.py
```diff
@@ -3340,35 +3340,35 @@ def setProcRam():
...
-    db_lib.setBeamlineConfigParam("fmx","spotNode6","cpu-004")        
-    db_lib.setBeamlineConfigParam("fmx","cbfComm","/usr/local/crys-local/bin/eiger2cbf-linux")
-    db_lib.setBeamlineConfigParam("fmx","dialsComm","/usr/local/crys-prod/phenix-1.11rc2-2531/build/bin/dials.find_spots_client")
+    db_lib.setBeamlineConfigParam("fmx","spotNode1","uranus-cpu007")
+    db_lib.setBeamlineConfigParam("fmx","spotNode2","uranus-cpu008")
+    db_lib.setBeamlineConfigParam("fmx","spotNode3","uranus-cpu009")
+    db_lib.setBeamlineConfigParam("fmx","spotNode4","uranus-cpu010")
+    db_lib.setBeamlineConfigParam("fmx","spotNode5","uranus-cpu008")
+    db_lib.setBeamlineConfigParam("fmx","spotNode6","uranus-cpu004")        
+    db_lib.setBeamlineConfigParam("amx","cbfComm","eiger2cbf")
```
RobertSchaffer1, line 3370: Should these two be amx?

lsdcGui.py
```diff
@@ -354,14 +354,14 @@ def __init__(self, parent = None, **kwargs):
     def getSpotNodeList(self):
       nodeList = []
       for i in range (0,self.spotNodeCount):
-        nodeList.append(int(getBlConfig("spotNode"+str(i+1)).split('-')[1]))
+        nodeList.append(int(getBlConfig("spotNode"+str(i+1)).split('-')[1].split('cpu')[1]))
```
RobertSchaffer1, line 357: Can you just drop the ".split('-')[1]" here? I think it would make the intent more clear, but I'm not sure if "cpu" would appear earlier in the string.

daq_macros.py
JunAishima: this is one of the functions that seems to be used for potentially re-setting up the config parameters, such as for a new beamline, or as examples of what to put in for a new beamline. I have merely replaced the hostnames with their new equivalents by adding "uranus-". there are amx equivalents for spotNodes a few lines above, lines 3354-3360.

lsdcGui.py
JunAishima: that is a good point, at least as long as they don't make "cpu" part of the first part of the name of  a computing node.

JunAishima: see commit 46341a6 below for the recommended fix (tested on beamline)

dalekreitler-bnl (approved):

RobertSchaffer1 (approved):

PR #144, opened 2021-11-17
Show pucks to be added by staff in chronological order

dalekreitler-bnl:
* reduce chance of loading incorrect pucks into dewar
* e.g. Puck-001, Puck001 will no longer be next to each other
* most recent batch of loaded pucks will be at top of list
* tested at amx


JunAishima (approved):
looks good! hopefully there will be scope for using the proposal number in the future

RobertSchaffer1 (approved):

PR #148, opened 2021-11-22
Use flyer for scan 141

JunAishima:
Implement new function `daq_macros.zebraDaqBluesky()` that sets up the flyer and performs the flyscan.

### DIFF
--- a/daq_lib.py
+++ b/daq_lib.py
@@ -698,7 +698,7 @@ def collect_detector_seq_hw(sweep_start,range_degrees,image_width,exposure_perio
     logger.info("vectorSync " + str(time.time()))    
     daq_macros.vectorSync()
     logger.info("zebraDaq " + str(time.time()))        
-    RE(daq_macros.zebraDaq(angleStart,range_degrees,image_width,exposure_period,file_prefix_minus_directory,data_directory_name,file_number,3,changeState))
+    daq_macros.zebraDaqBluesky(flyer,angleStart,range_degrees,image_width,exposure_period,file_prefix_minus_directory,data_directory_name,file_number,3,changeState))
 #    daq_macros.zebraDaq(angleStart,range_degrees,image_width,exposure_period,file_prefix_minus_directory,data_directory_name,file_number,3,protocol=protocol)  #?protocol?
   elif (protocol == "vector"):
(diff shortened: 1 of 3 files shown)


daq_macros.py
```diff
@@ -3098,6 +3098,38 @@ def zebraCamDaq(zebra,angle_start,scanWidth,imgWidth,exposurePeriodPerImage,file
   
   setPvDesc("lowMagTrigMode",0)
 
+def zebraDaqBluesky(flyer, angle_start, scanWidth, imgWidth, exposurePeriodPerImage, filePrefix, data_directory_name, file_number_start, scanEncoder=3, changeState=True):
```
RobertSchaffer1, line 3101: Is this entirely replacing the function below? If so, it would be good to remove it to clean up. If not, do we need to change other things in there, such as changing daq_lib.setRobotGovState() to gov_lib.setGovRobot()?

JunAishima: I think zebraDaq is being called from several places, so I only want to make this particular one use the new Bluesky one. I would like to make sure everything works, including where to put RE(), and sending references to the flyer object before I start cleaning up daq_macros.

daq_lib.py
```diff
@@ -698,7 +698,7 @@ def collect_detector_seq_hw(sweep_start,range_degrees,image_width,exposure_perio
     logger.info("vectorSync " + str(time.time()))    
     daq_macros.vectorSync()
     logger.info("zebraDaq " + str(time.time()))        
-    RE(daq_macros.zebraDaq(angleStart,range_degrees,image_width,exposure_period,file_prefix_minus_directory,data_directory_name,file_number,3,changeState))
+    daq_macros.zebraDaqBluesky(flyer,angleStart,range_degrees,image_width,exposure_period,file_prefix_minus_directory,data_directory_name,file_number,3,changeState))
```
RobertSchaffer1, line 701: I can't find flyer defined anywhere in this file, am I missing something?

RobertSchaffer1 (approved):

daq_lib.py
RobertSchaffer1: Resolved(ish) offline.

PR #167, opened 2022-01-11
Update conda envs for 2022 1

JunAishima:
Point to conda environments created for 2022-1 cycle 1. For both the server and gui envs, we have a symbolic link to -latest so that any updates and fixes to the environments don't require changing all references to specific environments.

Example: lsdc-server-2022-1.0 is a particular environment

- lsdc-server-2022-1-latest is the symbolic link to the current latest environment and used in LSDC code
- An environment lsdc-server-2022-1.1 can be created, and the -latest can be pointed to this new one, while the LSDC code does not need to be modified


mrakitin (approved):
Looks like a straightforward update.

PR #169, opened 2022-01-12
Fixes 2022 01 11

JunAishima:
Various fixes to robot and other code while I was getting LSDC server and GUI to run on the beamline on Tuesday.

### DIFF
--- a/gov_lib.py
+++ b/gov_lib.py
@@ -6,10 +6,10 @@ import logging
 logger = logging.getLogger(__name__)
 
 def set_detz_in(gov_robot, distance):
-    gov_robot.dev.dz.target_In.set(distance)
+    gov_robot.dev.dz.target_Work.set(distance)
 
 def set_detz_out(gov_robot, distance):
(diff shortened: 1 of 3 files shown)


gov_lib.py
```diff
@@ -6,10 +6,10 @@
 logger = logging.getLogger(__name__)
 
 def set_detz_in(gov_robot, distance):
```
mskinner5278, line 8: I think we might need to rename these functions, or make new functions for set_detz_work and set_detz_safe. Having set_detz_in set detz's work seems like it will come back later.

JunAishima: I would prefer we keep it for now, since we're just going into the cycle. however, we will need to keep an eye on it for AMX/FMX and make sure that the position names match between all beamlines.

if you like, you can put in a github issue for us to do for 2022 cycle 2, where we give it more logical names (based on function in this case) instead of just position (in/out). if you do, please make a comment and link to it (#<number>)

mskinner5278 (approved):

Issue #175, opened 2022-01-14
NYX robot move happening despite robot checked off


JunAishima:

JunAishima:
fixed as part of the #188 pull request. closing.

PR #176, opened 2022-01-18
put robot mounting/unmount moves in robot_online blocks

JunAishima:
* make sure calls to actual robot moves are blocked away within
   if getBlConfig('robot_online'): blocks
 * refs #175

### DIFF
--- a/denso_robot.py
+++ b/denso_robot.py
@@ -24,13 +24,14 @@ class DensoRobot:
         return MOUNT_STEP_SUCCESSFUL, kwargs
 
     def mount(self, gov_robot, puck_pos: int, pin_pos: int, samp_id: str, **kwargs):
-        try:
-            denso_puck_pos, denso_pin_pos = get_denso_puck_pin(puck_pos, pin_pos)
-            logger.info(f'Mounting: {denso_puck_pos} {denso_pin_pos}')
-            yield from self.robot.mount(denso_puck_pos, denso_pin_pos)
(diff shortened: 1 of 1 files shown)


JunAishima:
need to test before making a full pull request.

RobertSchaffer1:
Should probably put in an else block with some logging of why it's doing nothing.

JunAishima:
> related to this, the robot_lib code to handle status is broken - it should not be using return value from RE() to get its status as this is not necessarily the value that is returned from the function call. The return value from RE() is the status of the plan.
> 
> two possible approaches:
> 
>     * use the documented RunEngineResult object to extract the function return value
> 
>     * use the success or failure flags from the RE() call for making decisions. care would be needed to ensure that failures are correctly handled

OK - our LSDC is still using an old Bluesky, so we can't use RunEngineResult. have implemented a check after the plan to see whether a pin is mounted as expected. this will determine whether the step fails or succeeds.

need to think about how this will work with EMBL robot server in the future

JunAishima:
@RobertSchaffer1 - commit https://github.com/NSLS-II/lsdc/pull/176/commits/125c4058d8390d5651cd0e14c4bf07d99b2b9700 is an improvement made upon your suggestion

RobertSchaffer1 (approved):
Looks fine to me, in principle we could merge the mount and unmount since they only differ very slightly, but in practice this is probably just as good.

PR #179, opened 2022-01-31
Adds blconfig handling for configuring camera angles to send to C2C

mskinner5278:


daq_lib.py
```diff
@@ -755,9 +755,9 @@ def checkC2C_X(x,fovx): # this is to make sure the user doesn't make too much of
   return 1
   
 
-def center_on_click(x,y,fovx,fovy,source="screen",maglevel=0,jog=0,viewangle="beam"): #maglevel=0 means lowmag, high fov, #1 = himag with digizoom option, 
+def center_on_click(x,y,fovx,fovy,source="screen",maglevel=0,jog=0,viewangle=daq_utils.CAM_ANGLE_BEAM): #maglevel=0 means lowmag, high fov, #1 = himag with digizoom option,
```
JunAishima, line 758: this is a bad variable name. see what is in daq_utils

daq_utils.py
```diff
@@ -27,6 +27,11 @@
 soft_motor_list = []
 global screenYCenterPixelsLowMagOffset
 screenYCenterPixelsLowMagOffset = 58
+# Constants for use with C2C
+global CAMERA_ANGLE_BEAM,CAMERA_ANGLE_ABOVE, CAMERA_ANGLE_BELOW
```
JunAishima, line 31: if you think about it, there are several different ways to look at these angles - looking along the beam or upstream from the detector, for CAMERA_ANGLE_BEAM. then for the ABOVE/BELOW, you can imagine top is either downstream or upstream (downstream-> away from ring). you should clarify this.

lsdcGui.py
```diff
@@ -4280,12 +4280,21 @@ def pixelSelect(self, event):
...
+          current_viewangle = daq_utils.mag2ViewAngle
+        elif (self.zoom3Radio.isChecked()): 
+          current_viewangle = daq_utils.mag3ViewAngle
+        elif (self.zoom4Radio.isChecked()):
+          current_viewangle = daq_utils.mag4ViewAngle
+        
         if (self.threeClickCount > 0): #3-click centering
           self.threeClickCount = self.threeClickCount + 1
-          comm_s = 'center_on_click(' + str(correctedC2C_x) + "," + str(correctedC2C_y) + "," + str(fov["x"]) + "," + str(fov["y"]) + "," + '"screen",jog=90)'          
+          comm_s = 'center_on_click(' + str(correctedC2C_x) + "," + str(correctedC2C_y) + "," + str(fov["x"]) + "," + str(fov["y"]) + "," + '"screen",jog=90,viewangle=' + str(current_viewangle) + ')'
```
JunAishima, line 4295: would suggest using f strings as the server is Python 3 already. adding that last parameter just indicates how much easier it could be made with f'{abc}'

only change parts that we touch as we go on - we can't afford to change all of LSDC to f strings yet.

JunAishima (changes requested):
some clarifying and debugging is necessary - please go through these change requests.

daq_utils.py
mskinner5278: I had thought about those other angles, but I decided to just focus on 3 simple ones to make sure it was all in place. I will expand the existing comments to clarify the specific orientation of each of these angles. We can add additional angles easily if you need them.

lsdcGui.py
mskinner5278: I will be very happy to change this to an f string.

lsdcGui.py
```diff
@@ -4280,12 +4280,21 @@ def pixelSelect(self, event):
...
+        elif (self.zoom4Radio.isChecked()):
+          current_viewangle = daq_utils.mag4ViewAngle
+        
         if (self.threeClickCount > 0): #3-click centering
           self.threeClickCount = self.threeClickCount + 1
-          comm_s = 'center_on_click(' + str(correctedC2C_x) + "," + str(correctedC2C_y) + "," + str(fov["x"]) + "," + str(fov["y"]) + "," + '"screen",jog=90)'          
+          comm_s = f'center_on_click({correctedC2C_x},{correctedC2C_y},{fov["x"]},{fov["y"]},"screen",jog=90,viewangle={current_viewangle})'          
         else:
-          comm_s = 'center_on_click(' + str(correctedC2C_x) + "," + str(correctedC2C_y) + "," + str(fov["x"]) + "," + str(fov["y"])  + "," + '"screen",0)'
+          comm_s = f'center_on_click({correctedC2C_x},{correctedC2C_y},{fov["x"]},{fov["y"]},"screen",0,viewangle={current_viewangle})'
```
JunAishima, line 4297: the optional parameters should all be specified as keyword-value pairs. otherwise, it is hard to know which keyword the "screen" and 0 refer to.

This is not just your issue (you added `viewangle=`), but please fix

JunAishima (changes requested):

daq_utils.py
JunAishima: ok - Kevin just verified that your description sounds correct. resolving

Issue #182, opened 2022-02-03
NYX fix robot not updating sample status successfully


JunAishima:
using the RunEngine to execute the robot sample mount/dismount plans means that the status flags used are not being passed back to robot_lib. make a new function that checks whether the sample is mounted/unmounted as expected

PR #183, opened 2022-02-03
fix robot not updating status after sample load/unload

JunAishima:
* check that sample is mounted/unmounted as we expect after a mount/
   unmount is performed
 * replaces incorrect use of value from RE() - that returns a tuple of runIDs of
   plans executed
 * refs #182

### DIFF
--- a/denso_robot.py
+++ b/denso_robot.py
@@ -73,7 +73,7 @@ class DensoRobot:
             check_occupied = 0
         if int(self.robot.spindle_occupied_sts.get()) == check_occupied and \
            int(self.robot.puck_num_sel.get()) == puck_pos - 1 and \
-           int(self.robot.sample_num_sel.get()) == pin_pos:
+           int(self.robot.sample_num_sel.get()) == pin_pos:  # make sure puck number and sample number coming from robot and LSDC are zero- or one-indexed as necessary
             if mount:
                 return MOUNT_SUCCESSFUL
             else:
--- a/denso_robot.py
+++ b/denso_robot.py
@@ -65,3 +65,21 @@ class DensoRobot:
 
     def finish(self):
         ...
+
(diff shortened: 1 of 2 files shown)


denso_robot.py
```diff
@@ -65,3 +65,21 @@ def unmount(self, gov_robot, puck_pos: int, pin_pos: int, samp_id: str):
...
     def finish(self):
         ...
+
+    def check_sample_mounted(self, mount, puck_pos, pin_pos):  # is the correct sample present/absent as expected during a mount/unmount?
+        if mount:
+            check_occupied = 1
+        else:
+            check_occupied = 0
+        if int(self.robot.spindle_occupied_sts.get()) == check_occupied and \
+           int(self.robot.puck_num_sel.get()) == puck_pos - 1 and \
```
RobertSchaffer1, line 75: Is this the zero indexed vs. one indexed issue?

JunAishima: yes - could add comments in that direction for all of the comparisons

mskinner5278 (approved):

RobertSchaffer1:
Comment above:

I don't understand the TODO here, isn't that what this line is doing?

JunAishima:
clarifying, there is a lot of status info within the EMBL robot code, which currently is just assumed to be successful. RE() does not return the function call value, so the various status codes need to be re-determined outside of RE() or replaced with a similar check as DensoRobot.

RobertSchaffer1:
So shouldn't it be "TODO: don't assume embl robot is successful" or something like that?

JunAishima:
I would update it, but already merged - it's more like "TODO: implement status when installed on AMX/FMX"

RobertSchaffer1:
Hm, alright, something to remember then.

PR #184, opened 2022-02-04
updates to sample alignment systems (C2C / focus) to allow for proper control from both off-axis and beam camera views

mskinner5278:


daq_utils.py
```diff
@@ -41,7 +46,7 @@ def setBlConfig(param, value, beamline=beamline):
...
+  global beamline,detector_id,mono_mot_code,has_beamline,has_xtalview,xtal_url,xtal_url_small,unitScaling,sampleCameraCount,xtalview_user,xtalview_pass,det_type,has_dna,beamstop_x_pvname,beamstop_y_pvname,camera_offset,det_radius,lowMagFOVx,lowMagFOVy,highMagFOVx,highMagFOVy,lowMagPixX,lowMagPixY,highMagPixX,highMagPixY,screenPixX,screenPixY,screenPixCenterX,screenPixCenterY,screenProtocol,screenPhist,screenPhiend,screenWidth,screenDist,screenExptime,screenWave,screenReso,gonioPvPrefix,searchParams,screenEnergy,detectorOffline,imgsrv_host,imgsrv_port,beamlineComm,primaryDewarName,lowMagCamURL,highMagZoomCamURL,lowMagZoomCamURL,highMagCamURL,owner,dewarPlateMap,mag1ViewAngle,mag2ViewAngle,mag3ViewAngle
```
RobertSchaffer1, line 49: Is the omission of mag4ViewAngle intentional? Also, I hate everything about this line...

daq_utils.py
```diff
@@ -32,6 +32,11 @@
 CAMERA_ANGLE_BEAM = 0 # viewing angle is in line with beam, upstream from the sample, facing downstream, top toward ceiling
 CAMERA_ANGLE_ABOVE = 1 # viewing angle is directly above sample facing downward, top of view is downstream
 CAMERA_ANGLE_BELOW = 2 # viewing angle is directly below sample facing upward, top of view is downstream
+global mag1ViewAngle, mag2ViewAngle, mag3VivewAngle, mag4ViewAngle
+mag1ViewAngle = CAMERA_ANGLE_BEAM
+mag2ViewAngle = CAMERA_ANGLE_BEAM
+mag3ViewAngle = CAMERA_ANGLE_BEAM
+mag4ViewAngle = CAMERA_ANGLE_BEAM
```
RobertSchaffer1, line 39: These shouldn't all be CAMERA_ANGLE_BEAM, should they?

lsdcGui.py
```diff
@@ -3677,9 +3677,25 @@ def omegaTweakPosCB(self):
 
     def focusTweakCB(self,tv):
       tvf = float(tv)*daq_utils.unitScaling        
+
+      current_viewangle = daq_utils.mag1ViewAngle
+      if (self.zoom2Radio.isChecked()):
```
RobertSchaffer1, line 3682: Meta question - why are these "Radio"?

daq_utils.py
JunAishima: this is a good set of defaults, that they are all along the beam, like on AMX/FMX. this also won't require any additional changes on AMX/FMX

daq_utils.py
JunAishima: agree with hating - lots of other things to fix before this though

daq_lib.py
```diff
@@ -761,6 +761,7 @@ def center_on_click(x,y,fovx,fovy,source="screen",maglevel=0,jog=0,viewangle=daq
   if (getBlConfig('robot_online')): #so that we don't move things when robot moving?
     robotGovState = (getPvDesc("robotSaActive") or getPvDesc("humanSaActive"))
     if (not robotGovState):
+      logger.info("Bad governor state, must be in SA to use C2C, aborting C2C movement")
```
JunAishima, line 764: logger.error would be better - will be more noticeable in logs

lsdcGui.py
JunAishima: you have only one view so selecting between the 4 zoom levels (Mag 1-4) makes sense. is that what you are asking?

also, the concept of buttons of which only one can be selected at a time has been called "radio buttons" for a long time - at least since the early HTML days

lsdcGui.py
```diff
@@ -4035,9 +4051,13 @@ def eraseDisplayCB(self): #use this for things like zoom change. This is not the
 
     def getCurrentFOV(self):
       fov = {"x":0.0,"y":0.0}
-      if (self.zoom2Radio.isChecked()):  #lowmagzoom      
-        fov["x"] = daq_utils.lowMagFOVx/2.0
-        fov["y"] = daq_utils.lowMagFOVy/2.0
+      if (self.zoom2Radio.isChecked()):  #lowmagzoom
+        if (daq_utils.sampleCameraCount==2):   #this is a hard assumption that when there are 2 cameras the second uses highmagfov   
+          fov["x"] = daq_utils.highMagFOVx
```
JunAishima, line 4056: this will need to be documented once we have that going - will be very easy to get tripped up. along with the camera URL names, etc. are all based on two low mag views and two high mag views.

the comment there is appropriate and useful.

worth thinking about how three cameras should be handled

lsdcGui.py
```diff
@@ -5269,7 +5289,10 @@ def initCallbacks(self):
       self.omega_pv = PV(daq_utils.motor_dict["omega"] + ".VAL")
       self.omegaTweak_pv = PV(daq_utils.motor_dict["omega"] + ".RLV")
       self.sampyTweak_pv = PV(daq_utils.motor_dict["sampleY"] + ".RLV")
-      self.sampzTweak_pv = PV(daq_utils.motor_dict["sampleZ"] + ".RLV")            
+      if (daq_utils.beamline=="nyx"): 
+        self.sampzTweak_pv = PV(daq_utils.motor_dict["sampleX"] + ".RLV")
```
JunAishima, line 5293: I thought that because we knew the axes were different, we would have put the Z axis into sampleX in the epx.db file. the config change would have reduced the need for this change.

this could get messy though, we would need to look through the code carefully for all places sample x/y/z are used

JunAishima (changes requested):

PR #189, opened 2022-02-09
Initial docs

JunAishima:
initial build of docs, with a single index page.

confirmed by building on my desktop and viewing the html.


RobertSchaffer1 (approved):
LGTM - I've already looked at most of it from when I was going to add to it. As you said, we'll have to test it in production.

PR #192, opened 2022-02-14
Center marker color select file submenu

mskinner5278:
quick test of one of the options for allowing the center marker to have a swappable color


RobertSchaffer1:
It'd be nice to see a demo of this, in case we have stylistic suggestions. Code looks good.

JunAishima:
this needs a coordinated discussion with AMX/FMX before it can be merged into the main LSDC.

JunAishima (commented):
can you work with Martin to get code with the menu in a place that he is happy with? from what I remember, he wanted it to be out of the way so people don't accidentally use it.

also, I don't want you to go down a rabbithole of restructuring the whole GUI, just something simple and straightforward. please paste in text of what is requested so we can see that the code reflects what is requested.

JunAishima (commented):
it took me a couple of readthroughs to understand why you went from crosshair to overlay. things like this are useful to explain in commit messages

JunAishima (commented):
the last commit seems to be an empty whitespace change. did you mean to keep that?

JunAishima:
to be more specific, https://github.com/NSLS-II/lsdc/pull/192/commits/d5d84200d1e299b9b7c2ded4cae917c4e8f6e14b is only whitespaces - should have provided an ID to make it more clear exactly which commit I meant

mskinner5278:
That is weird, it looks like there was originally conflict and I got rid of the rest of the changes. I'm pretty sure the changes originally made by that commit were overwritten by a later one, but I would have expected it to still show those.

JunAishima (approved):

PR #195, opened 2022-02-18
skip expensive comparison if only one request on sample

JunAishima:
* that request would be the current one
 * if multiple requests, need to be more careful, so do
   the slow comparison

Practically speaking, this was able to cut down the amount of time at the beamline from several seconds down to tens of ms according to @dalekreitler-bnl .

### DIFF
--- a/daq_macros.py
+++ b/daq_macros.py
@@ -3386,12 +3386,14 @@ def setDimpleCommand(commName):
 def lastOnSample():
   if (ednaActiveFlag == 1):
     return False
-  r = db_lib.popNextRequest(daq_utils.beamline)
-  if (r != {}):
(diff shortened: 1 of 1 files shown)


RobertSchaffer1 (approved):

dalekreitler-bnl:
When testing by loading pucks into the DewarTree and running: %timeit lastOnSample() in the ipython console, servers that include this commit have a 10x speedup (~300 ms vs. ~30 ms). Also the delay no longer scales with puck number.

Further tested by running automated collection on one puck (with 24 loaded in DewarTree), so that there were 3 completed requests for each sample; this didn't seem to make much difference for 1 puck at least. The %timeit lastOnSample() test was about 35 ms either way (with or w/o completed requests in the queue). I think we agreed to merge this one and think of a good long term solution for later.

dalekreitler-bnl (approved):

PR #198, opened 2022-02-24
Final Day1 NYX GUI fixes

mskinner5278:
included:
- Fake gripper temp replaced with N/A
- Spreadsheet importing "" file handling at server level
- sample file number start field validation at GUI level, allows positive integers 1-99999
- User Extras button disabling for NYX

### DIFF
--- a/parseSheet.py
+++ b/parseSheet.py
@@ -63,9 +63,11 @@ def insertSpreadsheetDict(d,owner):
 
 
 def importSpreadsheet(infilename,owner):
-  d = parseSpreadsheet(infilename)
-  insertSpreadsheetDict(d,owner)
-
+  if(infilename != ""):
+    d = parseSpreadsheet(infilename)
+    insertSpreadsheetDict(d,owner)
+  else:
+    logger.error("import spreadsheet error:  no filename given")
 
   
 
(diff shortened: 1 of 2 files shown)


parseSheet.py
```diff
@@ -63,9 +63,11 @@ def insertSpreadsheetDict(d,owner):
 
 
 def importSpreadsheet(infilename,owner):
-  d = parseSpreadsheet(infilename)
-  insertSpreadsheetDict(d,owner)
-
+  if(infilename != ""):
```
JunAishima, line 66: this should handle more than just an explicit empty string from a cancel - it should also handle empty or whitespace entry as well. see the best answer for https://stackoverflow.com/questions/9573244/how-to-check-if-the-string-is-empty

JunAishima (changes requested):

parseSheet.py
JunAishima: you have a very good point, and are getting towards the best solution.

since the goal is either good data from the spreadsheet or asking the user to select a valid file, you should probably check for not just an OSError for a missing/unreadable file, but also for badly formatted file.

I would suggest checking offline to see what exception that throws.

minimize the types caught to prevent other issues to prevent other issues from being missed

JunAishima (approved):
looks good - those error messages will be useful for users (and us) to diagnose their issues. it's easy to select the wrong file

parseSheet.py
JunAishima: changes look good! already approved, so resolving.

PR #199, opened 2022-02-28
One branch to rule them all

JunAishima:
This branch is a combination of master (descendant of current AMX/FMX code), scans-ophyd-bluesky (FMX Bluesky work), and nyx_ophyd (NYX Bluesky work).

it will be used as the base for future nyx_ophyd to continue Bluesky development for experiments on NYX, and for adapting the combined code for use on AMX/FMX.

Currently a draft request that is more informational for what the branch is meant to be used for than necessarily one to be merged into master.


RobertSchaffer1:
I'll start (likely quite slowly!) going through this, is there any particular part I should look at first?

beamline_lib.py
```diff
@@ -6,6 +6,8 @@
 import beamline_support
 import daq_lib
 import db_lib
+import gov_lib
```
RobertSchaffer1, line 9: Probably don't need this, based on the commented code below?

config_params.py
```diff
@@ -48,3 +52,24 @@ class RasterStatus(Enum):
...
+MOUNT_UNRECOVERABLE_ERROR = 2
+MOUNT_STEP_SUCCESSFUL = 3
+
+UNMOUNT_FAILURE = 0
+UNMOUNT_SUCCESSFUL = 1
+UNMOUNT_STEP_SUCCESSFUL = 2
+
+PINS_PER_PUCK = 16
+
+DETECTOR_OBJECT_TYPE_LSDC = "lsdc"  # using det_lib
```
RobertSchaffer1, line 66: I feel like these three should go above somewhere?

JunAishima:
> I'll start (likely quite slowly!) going through this, is there any particular part I should look at first?

you have probably seen a lot of the code by now, so the merge points are the most critical to make sure I didn't make any mistakes when bringing master+scans_ophyd_bluesky together, and then the merge with nyx_ophyd.

Remember that you can see the branch history graphically by looking at github.com -> LSDC -> Insights -> Network

RobertSchaffer1:
> > I'll start (likely quite slowly!) going through this, is there any particular part I should look at first?
> 
> you have probably seen a lot of the code by now, so the merge points are the most critical to make sure I didn't make any mistakes when bringing master+scans_ophyd_bluesky together, and then the merge with nyx_ophyd.
> 
> Remember that you can see the branch history graphically by looking at github.com -> LSDC -> Insights -> Network

Makes a lot of sense... hmm, I'm really not sure what the best way to go about doing this is, look only at files that have been modified by multiple branches maybe? The issue with that approach is that a lot of the files are massive, and those are the ones that have been changed the most.

mskinner5278 (changes requested):

JunAishima:
this draft is not too useful - especially now that things have moved on. as we have been reviewing code that goes into both master and one-branch, the merge should not be too difficult (though needs to be tested).

closing.

PR #212, opened 2022-04-05
Server state display

mskinner5278:
The plan is to create a heartbeat PV that the server constantly updates with a random number. If the server finds that something else has changed that number, it will report that there are multiple servers running. If the GUI finds that the number has not changed for more than 5~ seconds, then it will report the server has shut down. All of these reports will go through the existing program_state pv.


mskinner5278:
Closing this for now to review the process

PR #222, opened 2022-06-14
MX queue collect input blocking

mskinner5278:
mx subpatch 2
* Queue Collect checkbox now switches selected sample instead of blocking user input

### DIFF
--- a/lsdcGui.py
+++ b/lsdcGui.py
@@ -4391,9 +4391,8 @@ class ControlMain(QtWidgets.QMainWindow):
           self.selectedSampleID = itemData
           if (getBlConfig("queueCollect") == 0):
             if (self.mountedPin_pv.get() != self.selectedSampleID):                    
+              self.selectedSampleID = self.mountedPin_pv.get()
               self.popupServerMessage("You can only add requests to a mounted sample, for now.")
-              self.progressDialog.close()              
-              return
 
         try:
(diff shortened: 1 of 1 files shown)


lsdcGui.py
```diff
@@ -4391,9 +4391,8 @@ def addRequestsToAllSelectedCB(self):
           self.selectedSampleID = itemData
           if (getBlConfig("queueCollect") == 0):
             if (self.mountedPin_pv.get() != self.selectedSampleID):                    
+              self.selectedSampleID = self.mountedPin_pv.get()
```
JunAishima, line 4394: is this meant to always queue a collection (because the return is removed), but always leave the sample selected?

mskinner5278: Correct, they felt it was unnecessarily stopping operations every time it happened, so the return was removed.

JunAishima: ok, looks strange, but the beamlines asked for it so I'm ok with it.

lsdcGui.py
```diff
@@ -4427,8 +4426,8 @@ def addSampleRequestCB(self,rasterDef=None,selectedSampleID=None):
 
       if (getBlConfig("queueCollect") == 0):
         if (self.mountedPin_pv.get() != self.selectedSampleID):                    
+          self.selectedSampleID = self.mountedPin_pv.get()
           self.popupServerMessage("You can only add requests to a mounted sample, for now.")
```
JunAishima, line 4430: Did Edwin want these popup messages to be removed? It seems like the behavior is different now, so it should either not be displayed at all or show something different.

PR #227, opened 2022-07-27
Gripper unmount cold checkbox

mskinner5278:
Only available when QueueEnable is checked on staff screen

### DIFF
--- a/lsdcGui.py
+++ b/lsdcGui.py
@@ -220,10 +220,17 @@ class StaffScreenDialog(QFrame):
           self.beamCheckOnCheckBox.setChecked(False)
         self.beamCheckOnCheckBox.stateChanged.connect(self.beamCheckOnCheckCB)
 
+        self.gripperUnmountColdCheckBox = QCheckBox("Unmount Cold")
+        self.gripperUnmountColdCheckBox.setChecked(False)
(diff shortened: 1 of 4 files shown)


lsdcGui.py
```diff
@@ -220,10 +220,17 @@ def __init__(self, parent = None, **kwargs):
           self.beamCheckOnCheckBox.setChecked(False)
         self.beamCheckOnCheckBox.stateChanged.connect(self.beamCheckOnCheckCB)
 
+        self.gripperUnmountColdCheckBox = QCheckBox("Unmount Cold")
+        self.gripperUnmountColdCheckBox.setChecked(False)
+        self.gripperUnmountColdCheckBox.setEnabled(False)
```
JunAishima: so unmount cold is only available with queue collect, and its default is true for whenever queue collect is on?

JunAishima (commented):
I think this is a good change, because it is not forcing all GUIs to turn off unmount cold when one starts up.

however, shouldn't beam_check be checked here before setting the state of the GUI and its enabled status?

lsdcGui.py
mskinner5278: The default should be false when queue collect is on.

PR #229, opened 2022-08-10
Nyx 2022 2

JunAishima:
These are the changes necessary to get LSDC running for this cycle.

The branch name on the NYX LSDC beamline repo is called conda-env-2022-2.3-update, but will be renamed here to refer to NYX and the cycle.

### DIFF
--- a/lsdcServer
+++ b/lsdcServer
@@ -1,3 +1,3 @@
 dzdo pkill -KILL -f daq_main2.py
-source /opt/conda_envs/lsdc-server-2022-1-latest/bin/activate
+source /opt/conda_envs/lsdc-server-2022-2.1/bin/activate
 $LSDCHOME/daq_main2.py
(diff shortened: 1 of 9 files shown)


lsdcServer
```diff
@@ -1,3 +1,3 @@
 dzdo pkill -KILL -f daq_main2.py
-source /opt/conda_envs/lsdc-server-2022-1-latest/bin/activate
```
mskinner5278, line 2: The one in daq_mainAux is 2.2 and this one is 2.3. Should we go back to using "latest" to solve this and prevent code updates for conda env changes?

JunAishima: let's just keep it as specific versions (2.2/2.3) for now. we may have a better deployment method for conda envs that creates the symlink automatically in the future, which would help us automate things more. I don't like having to make the symlink manually as it is right now

PR #230, opened 2022-08-10
Removed progress bar at startup

vshekar:
By calling the reset() slot on the QProgressDialog, the dialog is hidden and can be reused. According to: https://doc.qt.io/qtforpython-5/PySide2/QtWidgets/QProgressDialog.html#PySide2.QtWidgets.PySide2.QtWidgets.QProgressDialog.reset


mskinner5278 (approved):

lsdcGui.py
```diff
@@ -5166,8 +5166,9 @@ def initUI(self):
         self.immediate_comm_pv = PV(daq_utils.beamlineComm + "immediate_command_s")
         self.stillModeStatePV = PV(daq_utils.pvLookupDict["stillModeStatus"])        
         self.progressDialog = QtWidgets.QProgressDialog()
-        self.progressDialog.setCancelButtonText(QString())
+        self.progressDialog.setCancelButtonText(QString("Cancel"))
```
JunAishima, line 5169: ideally, could you make this a separate commit next time? also, was this done so you could help diagnose what it was?

vshekar: Correct, It was done to verify I was modifying the correct dialog box. I will put something like this in a separate commit next time

PR #233, opened 2022-09-01
Extract detector metadata

JunAishima:
The LSDC side of the mxtools changes to make rastering run in LSDC. Some notes:

- I wasn't able to get metadata handling to be correct so have mostly put in the bare essential for the code to run. It won't return anything useful
- The detector is handled via function calls, but the vector scan is run as fly scans. This may contribute to its complexity.
- The rastering looks good and the number of images collected agrees with expectations, however it has not been tested with beam to verify that the images are being collected correctly.

### DIFF
--- a/daq_macros.py
+++ b/daq_macros.py
@@ -1921,7 +1921,7 @@ def snakeRasterBluesky(rasterReqID, grain=""):
       db_lib.updateRequest(rasterRequest)
       daq_lib.set_field("xrecRasterFlag",rasterRequest["uid"])
       logger.info(f'setting xrecRasterFlag to: {rasterRequest["uid"]}')
-    raster_flyer.detector.cam.acquire.put(0, wait=True)
+    yield from raster_flyer.detector.collect()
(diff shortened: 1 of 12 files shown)


daq_macros.py
```diff
@@ -1655,12 +1715,13 @@ def reprocessRaster(rasterReqID):
   procFlag = 1
   
   for i in range(len(rasterDef["rowDefs"])):
+    currentRow = rasterDef["rowDefs"][i]
     time.sleep(0.5)
```
mskinner5278, line 1719: Why is this sleep in the middle of a processing loop?

mskinner5278 (changes requested):

daq_macros.py
JunAishima: I am guessing this is to prevent all of the row reprocessing from being dumped onto processing nodes at once, which would overwhelm them. fortunately or unfortunately, this is not one of the parts that I changed...

mskinner5278 (approved):

PR #234, opened 2022-09-06
[New] Added DewarRefill function to lsdcServer

mskinner5278:
* Accepts HOURS as input
1. Calls DewarAutoFillOff() and DewarHeaterOn()
2. Waits for x hours and then calls DewarAutoFillOn() and DewarHeaterOn()
 * Also added DewarRefillCancel() function to lsdcServer
 * * Please wait for the confirmation (60 seconds maximum) before attempting to start a new timer

### DIFF
--- a/robot_lib.py
+++ b/robot_lib.py
@@ -155,6 +155,10 @@ def dryGripper():
 def DewarRefill(hours):
   global _dewarRefillThread
   seconds = int((hours * 60 * 60))
+  if _dewarRefillThread is not None:
+    if _dewarRefillThread.is_alive():
+      logger.info("An existing DewarRefillTask is already running.")
+      return
   _dewarRefillThread = Thread(target=_dewarRefillTask, args=(seconds,0))
   _dewarRefillThread.start()
 
--- a/robot_lib.py
+++ b/robot_lib.py
@@ -36,6 +36,7 @@ sampZadjust = 0
 global retryMountCount
 retryMountCount = 0
 
+
 def finish():
   if (getBlConfig('robot_online')):  
     try:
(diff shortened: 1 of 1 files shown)


robot_lib.py
```diff
@@ -151,6 +152,37 @@ def dryGripper():
...
+  goal_time = time.time() + (seconds)
+  while not dewarRefillStop:
+    if time.time() < goal_time:
+      time_remaining = round(((goal_time - time.time()) / 60), 1)
+      logger.info(f"DewarRefillTask: Time remaining until auto fill on... {time_remaining} minutes")
+      time.sleep(60)
+    else:
+      logger.info("Dewar refill task running.")
+      DewarAutoFillOn()
+      DewarHeaterOn()
```
JunAishima, line 176: is this supposed to be `DewarHeaterOff()`? also, should this be called if cancel is called?

robot_lib.py
```diff
@@ -151,6 +152,37 @@ def dryGripper():
     daq_lib.gui_message("Dry gripper failed! " + e_s)
     setPvDesc("warmupThreshold",saveThreshold)          
     
+def DewarRefill(hours):
+  global _dewarRefillThread
+  seconds = int((hours * 60 * 60))
+  _dewarRefillThread = Thread(target=_dewarRefillTask, args=(seconds,0))
+  _dewarRefillThread.start()
+
+def _dewarRefillTask(seconds, flag):
```
JunAishima, line 161: `flag` is not used so should be removed

robot_lib.py
```diff
@@ -151,6 +152,37 @@ def dryGripper():
     daq_lib.gui_message("Dry gripper failed! " + e_s)
     setPvDesc("warmupThreshold",saveThreshold)          
     
+def DewarRefill(hours):
+  global _dewarRefillThread
+  seconds = int((hours * 60 * 60))
+  _dewarRefillThread = Thread(target=_dewarRefillTask, args=(seconds,0))
```
JunAishima, line 158: could you check for the existence of the thread to prevent it being called multiple times? otherwise, you might get the function being called multiple times and having the timer being reset

JunAishima (commented):
can you add text explaining why this is being done? it turns off the auto refill to be resumed some hours afterwards - is the refill bad, requiring this toggling? should the heater function be turned off/on instead of just being turned on?

robot_lib.py
mskinner5278: I asked that specifically when we tested it. It is supposed to be on for both.

robot_lib.py
mskinner5278: Fixed with 61ad2c5ffb84804c3b0f23ff3245f9952b4c5fc0

robot_lib.py
mskinner5278: Fixed with 3e738822965449fc5567ad4a09ee9eb5b17e1f61

JunAishima (approved):
looks good now!

mskinner5278:
This is being added to automated a process that is typically performed by staff:

Apparently ~once a week, a staff member wakes up early and turns on the dewar heater and autofill function. This new server command will let them set a delay timer for when they want these functions to automatically trigger. They can call DewarRefill(12) at 6 pm and the process will happen automatically at 6 am the following morning. If it needs to be canceled, DewarRefillCancel() issues a trip for the thread's loop and halts it.

JunAishima:
> 

Is this done because the dewar heating and autofill are turned off after the last set of users, so need to be turned on to prepare for the next users when there is a gap between them?

PR #235, opened 2022-09-06
Seperated hutch cameras to own thread + FPS inc

vshekar:
Allows LSDC to start without hutch camera feeds, and camera feeds can have faster frame rate (currently set to 30 FPS)

### DIFF
--- a/lsdcGui.py
+++ b/lsdcGui.py
@@ -38,7 +38,7 @@ from element_info import element_info
 import numpy as np
 import _thread #TODO python document suggests using threading! make this chance once stable
 import lsdcOlog
-
+from threads import VideoThread
 import socket
 hostname = socket.gethostname()
 ws_split = hostname.split('ws')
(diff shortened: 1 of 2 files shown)


mskinner5278 (approved):

JunAishima (commented):
a couple of general questions:

- does this make everything about self.timerHutch and self.timerSample obsolete? if so, we should really clean up all of the other code that uses them.
- have you tested this and tried different delays?

vshekar:
> a couple of general questions:
> 
>     * does this make everything about self.timerHutch and self.timerSample obsolete? if so, we should really clean up all of the other code that uses them.
> 
>     * have you tested this and tried different delays?

* This specific PR makes only self.timerHutch obsolete, timerSample should be similar fix. I can add it to this PR or make a separate one. I agree, references to timerHutch should be removed and I can remove them. Currently self.timerHutch doesn't trigger a function when it times out.
* I have tested this in my local environment and on AMX. Tested original 1 FPS, 24 FPS and 30 FPS

JunAishima:
> * This specific PR makes only self.timerHutch obsolete, timerSample should be similar fix. I can add it to this PR or make a separate one. I agree, references to timerHutch should be removed and I can remove them. Currently self.timerHutch doesn't trigger a function when it times out.

Could you try to clean up timerHutch and timerSample in this same pull request? It would make a nice, connected piece of work that will make a noticeable difference to users.

JunAishima (approved):
looks good now

PR #237, opened 2022-09-10
Pucks list sorted in desc order of modification

vshekar:
Whenever a new puck is created or an existing puck is updated, db_lib creates/updates a field called "modified_time" to the current time. In the GUI these pucks are ordered in descending order of modified_time If no modified_time exists, it's modified_time is considered 0 for sorting purposes


JunAishima (approved):
looks good, have seen it working as well.

PR #241, opened 2022-11-01
Vector tweak changes

vshekar:
Set of changes that updates a vectors start and end points correctly when they are moved in a 2D plane

### DIFF
--- a/daq_utils.py
+++ b/daq_utils.py
@@ -124,6 +124,28 @@ def wave2energy(w, digits=2):
   else:
     return float(f"%.{digits}f" % (EV_ANGSTROM_CONSTANT/w))
 
+def gonio2lab(x_gonio, y_gonio, z_gonio, omega_deg):
+  """Utility function to convert gonio co-ordinates to lab co-ordinates
+  """
+  #YL = cO*YP + sO*ZP 
+  #ZL = -sO*YP + cO*ZP  
+  cosO = math.cos(math.radians(omega_deg))
+  sinO = math.sin(math.radians(omega_deg))
+  y_lab = (cosO * y_gonio) + (sinO * z_gonio)
+  z_lab = -(sinO * y_gonio) + (cosO * z_gonio)
+  return x_gonio, y_lab, z_lab, omega_deg
+
+def lab2gonio(x_lab, y_lab, z_lab, omega_deg):
+  """Utility function to convert lab co-ordinates to gonio co-ordinates
+  """
+  #YP = cO*YL - sO*ZL
+  #ZP = sO*YL + cO*ZL
+  cosO = math.cos(math.radians(omega_deg))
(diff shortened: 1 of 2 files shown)


daq_utils.py
```diff
@@ -124,6 +124,28 @@ def wave2energy(w, digits=2):
   else:
     return float(f"%.{digits}f" % (EV_ANGSTROM_CONSTANT/w))
 
+def gonio2lab(x_gonio, y_gonio, z_gonio, omega_deg):
```
JunAishima, line 127: does this follow one of NSLS-II or beamline conventions for x, y, z?

lsdcGui.py
```diff
@@ -4714,30 +4714,90 @@ def removePuckCB(self):
       dewarPos, ok = DewarDialog.getDewarPos(parent=self,action="remove")
       
 
-    def getVectorObject(self):
+    def transform_vector_coords(self, prev_coords, current_raw_coords):
```
JunAishima, line 4705: good documentation of complex functions!

lsdcGui.py
```diff
@@ -4749,34 +4809,19 @@ def drawVector(self):
...
-      try:
-        if self.vecLine != None:
-          self.scene.removeItem(self.vecLine)
-      except AttributeError: # likely due to self.vecLine not defined yet
-        pass
-
+      if self.vecLine:
+        self.scene.removeItem(self.vecLine)
+        self.vecLine = None
+
```
JunAishima, line 4820: do we know whether we are still getting these `AttributeError`s or not? (several areas of this commit) we may get more crashes of the GUI

JunAishima (dismissed):
Generally, it is good to see some refactoring to make this code easier to read while adding the new functionality.

lsdcGui.py
vshekar: We shouldn't get attribute error for this specific variable as it is initialized to None: [here](https://github.com/NSLS-II/lsdc/blob/6dce1d9c5927b1b41394823e48526887458b8bd1/lsdcGui.py#L2203). So it should be safe to check its value anywhere in the app

daq_utils.py
vshekar: Unsure what the NSLS-II conventions are, I think it uses beamline conventions. X is the direction of the sample holder, Y is up and down and Z should be in the direction of the beam (sorry for the poor description)

lsdcGui.py
JunAishima: you're right, as long as it is defined, we shouldn't need that exception. resolving this.

daq_utils.py
JunAishima: now that I think about it, I think this should be added to the general documentation, not what you are working on.

resolving

PR #242, opened 2022-11-07
Isara ophyd object integration

mskinner5278:


start_bs.py
```diff
@@ -169,9 +169,9 @@ class SampleXYZ(Device):
     from mxtools.raster_flyer import MXRasterFlyer
     raster_flyer = MXRasterFlyer(vector, zebra, detector)
 
-    from nyxtools.robot import DensoOphydRobot
+    from nyxtools.isara_robot import IsaraRobotDevice
     from denso_robot import DensoRobot
-    denso_ophyd_robot = DensoOphydRobot("XF:19IDC-ES{Rbt:1}", name="robot")
+    denso_ophyd_robot = IsaraRobotDevice("XF19IDC-ES{Rbt:1}", name="robot")
```
JunAishima, line 174: if you can make an Isara or generic subclass of DensoRobot, then you will also be able to clean up this and the following line with appropriate names for the objects and classes

JunAishima (changes requested):

denso_robot.py
```diff
@@ -75,12 +83,17 @@ def unmount(self, gov_robot, puck_pos: int, pin_pos: int, samp_id: str):
     def finish(self):
         ...
 
+    def robotCanExchangeSamples():
```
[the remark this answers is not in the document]
mskinner5278: It is named this way for readability on the LSDC interface side, but it is for representing that the robot can accept mount commands when a sample is already mounted. I did not make the assumption that it was a feature exclusive to the doublegripper.

denso_robot.py
```diff
@@ -75,12 +83,17 @@ def unmount(self, gov_robot, puck_pos: int, pin_pos: int, samp_id: str):
...
+        return True
+
     def check_sample_mounted(self, mount, puck_pos, pin_pos):  # is the correct sample present/absent as expected during a mount/unmount?
         if getBlConfig('robot_online'):
             if mount:
                 check_occupied = 1
             else:
                 check_occupied = 0
+            puck_pos+=1
```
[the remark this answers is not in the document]
mskinner5278: denso puck/pin was 0 indexed, isara is 1 indexed. It needed to be handled at the time, but we can easily move it to a different location if needed.

start_bs.py
mskinner5278: I will make a more generic name for it.

denso_robot.py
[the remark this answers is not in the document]
JunAishima: what we need is to be able to read the code and understand why that was done. can you add some comments throughout this your module to note this difference?

start_bs.py
mskinner5278: added a more generic interface with 2b38bd0

denso_robot.py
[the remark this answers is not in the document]
mskinner5278: added the comments with 35f92e4

denso_robot.py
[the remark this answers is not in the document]
mskinner5278: in 1b27084 I changed this to a parameter which is declared in start_bs startup process, so it will no longer require modification of amx/fmx code.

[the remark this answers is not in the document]
mskinner5278: I added additional comments about it to the code in 1b27084

start_bs.py
```diff
@@ -125,6 +125,7 @@ class SampleXYZ(Device):
 
     from embl_robot import EMBLRobot
     robot = EMBLRobot()
+    robot.canExchangeSamples = False
```
JunAishima, line 128: you would need to make the EMBL robot an Ophyd robot or something to be able to get the `canExchangeSamples` variable

denso_robot.py
```diff
@@ -100,3 +110,8 @@ def check_sample_mounted(self, mount, puck_pos, pin_pos):  # is the correct samp
                     return UNMOUNT_FAILURE
         else:
             return MOUNT_STEP_SUCCESSFUL  # always successful if robot is not online
+
+# This is a generic interface to give robots access to the robot_lib api
+class OphydRobot(DensoRobot):
+    def __init__(self):
+        self.robot = robot
```
JunAishima, line 117: robot is not defined here - did you mean to pass in a robot?

but if you do that, then to access any functions, you would need to do something like `robot.robot.dismount()`.

did you actually mean to use a mixin or something similar so that you can add the canExchangeSamples in your class?

denso_robot.py
```diff
@@ -100,3 +110,8 @@ def check_sample_mounted(self, mount, puck_pos, pin_pos):  # is the correct samp
                     return UNMOUNT_FAILURE
         else:
             return MOUNT_STEP_SUCCESSFUL  # always successful if robot is not online
+
+# This is a generic interface to give robots access to the robot_lib api
+class OphydRobot(DensoRobot):
```
JunAishima, line 115: I'm not sure I understand what you mean to do here - you pass in a robot object, but also subclass DensoRobot?

start_bs.py
```diff
@@ -169,13 +171,13 @@ class SampleXYZ(Device):
...
-    denso_ophyd_robot = DensoOphydRobot("XF:19IDC-ES{Rbt:1}", name="robot")
-    robot = DensoRobot(denso_ophyd_robot) # DensoRobot is the robot_lib API-compatible object
+    from nyxtools.isara_robot import IsaraRobotDevice
+    from denso_robot import OphydRobot
+    ophyd_robot = IsaraRobotDevice("XF19IDC-ES{Rbt:1}", name="robot")
+    robot = OphydRobot(ophyd_robot) # OphydRobot is the robot_lib API-compatible object
     govs = _make_governors("XF:19IDC-ES", name="govs")
     gov_robot = govs.gov.Robot
-
+    robot.canExchangeSamples = True # necessary for us to identify double-gripper availability
```
JunAishima, line 180: the canExchangeSamples variable is not defined anywhere

JunAishima (changes requested):
I'm not sure you have the canExchangeSamples variable correct in this implementation. have you tested this on the beamline? you may want to redo this code and resubmit

start_bs.py
```diff
@@ -172,9 +172,9 @@ class SampleXYZ(Device):
     raster_flyer = MXRasterFlyer(vector, zebra, detector)
 
     from nyxtools.isara_robot import IsaraRobotDevice
-    from denso_robot import DensoRobot
-    denso_ophyd_robot = IsaraRobotDevice("XF19IDC-ES{Rbt:1}", name="robot")
-    robot = DensoRobot(denso_ophyd_robot) # DensoRobot is the robot_lib API-compatible object
+    from denso_robot import OphydRobot
+    ophyd_robot = IsaraRobotDevice("XF19IDC-ES{Rbt:1}", name="robot")
+    robot = OphydRobot(ophyd_robot) # DensoRobot is the robot_lib API-compatible object
```
JunAishima: you probably want to change of the class name in the comment

JunAishima (changes requested):

start_bs.py
mskinner5278: this was done in d8ce8bf

start_bs.py
mskinner5278: taken care of with 671207e

start_bs.py
mskinner5278: change added to embl code to handle the function required 671207e

denso_robot.py
```diff
@@ -100,3 +113,8 @@ def check_sample_mounted(self, mount, puck_pos, pin_pos):  # is the correct samp
...
```
JunAishima, line 118: is there any reason why you couldn't rename `DensoRobot` as `OphydRobot`? you haven't added any value by making the `OphydRobot` class

denso_robot.py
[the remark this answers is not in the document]
JunAishima: I still don't find the name descriptive, and this https://github.com/NSLS-II/lsdc/commit/1b270847bd1dab222056f1febceb4308a11c9f38 is no longer part of this branch.

would a name like `multiSampleGripper` be more appropriate?

when I see `robotCanExchangeSample`, it brings to mind that of course a sample exchange robot can exchange samples, not a name for a particular type of gripper.

JunAishima (changes requested):
the implementation is looking good, but I still have a couple of small points I would like you to fix before I approve.

I'm still not convinced by the `robotCanExchangeSamples` name (whose description has been lost on another branch), and the layering of the `OphydRobot` class. please justify or fix.

denso_robot.py
mskinner5278: renamed the base class and removed the interface in [c59660c](https://github.com/NSLS-II/lsdc/pull/242/commits/c59660cc9cbff23975eb43960b4703ed294bbacc)

denso_robot.py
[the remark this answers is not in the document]
mskinner5278: updated to your name in [2737f18](https://github.com/NSLS-II/lsdc/pull/242/commits/2737f181bebd985b2204daac7652ec9138553e1e)

JunAishima (approved):

PR #244, opened 2022-12-09
Bug fixes and enhancements

vshekar:


lsdcGui.py
```diff
@@ -878,8 +878,6 @@ def __init__(self,parent = None):
         vBoxDialsParams.addLayout(self.hBoxRasterLayout3)
         dialsGB.setLayout(vBoxDialsParams)
 
-        reprocessRasterButton = QtWidgets.QPushButton("ReProcessRaster")
```
JunAishima, line 881: there wasn't any interest in keeping it, but just re-doing it with dozor? I would have thought having this option would be useful enough to keep.

another issue is that getting rid of this gets rid of a lot of config parameters and code here and in the `lsdc-processing` repo. it may make more sense to comment out adding the button in LSDC, then make a separate pull request to completely clean things out, and another for lsdc-processing as well

JunAishima (changes requested):

lsdcGui.py
JunAishima: looks like my comment's suggestion is generally being followed (see https://github.com/NSLS-II/lsdc/pull/244/commits/b52bf7742411520d5771b60745011e21fd81d07b) so I am considering this resolved.

JunAishima (approved):
the changes are a good start to cleaning up the GUI in general, and modernizing the interaction users have with it. approve

PR #246, opened 2022-12-23
2022 3 amxfmx

JunAishima:
small changes done during the cycle

- puck ordering
- c3d location
- Edwin's MT_LoadCheck

### DIFF
--- a/bin/mt-load-check
+++ b/bin/mt-load-check
@@ -12,12 +12,12 @@ esac
 case $servorws in
 	ws*)
 		echo Workstation - activating GUI env
-		conda activate lsdc-gui-2022-1.1
+		PYTHONTOUSE=/opt/conda_envs/lsdc-gui-2022-1.1/bin/python
 		;;
 	*)
 		echo Server - activating server env
-		conda activate lsdc-server-2022-3.2
+		PYTHONTOUSE=/opt/conda_envs/lsdc-server-2022-3.2/bin/python
 		;;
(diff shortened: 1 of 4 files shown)


bin/mt-load-check
```diff
@@ -12,12 +12,12 @@ esac
 case $servorws in
 	ws*)
 		echo Workstation - activating GUI env
-		conda activate lsdc-gui-2022-1.1
+		PYTHONTOUSE=/opt/conda_envs/lsdc-gui-2022-1.1/bin/python
```
vshekar, line 15: An environment variable name like PYTHONENV or PYTHON_ENV may be more readable

JunAishima: I agree - looking back at this, at first glance, I saw it as PYTHONHOUSE!

JunAishima: would you be happy with PYTHON_BINARY or CONDA_ENV_PYTHON_BINARY? or is there something else that you think would be better?

JunAishima: ok, going with `CONDA_ENV_BASE`, and bin/python will be appended to the directory name to define the python binary location

vshekar (approved):

JunAishima:
ok, now I understand what happened... I had made the changes onto NSLS-II/2022-3-amxfmx but then never merged that branch into master. that's why I remembered these changes having been approved in a pull request.

PR #248, opened 2022-12-23
Fix raster 2022 11

JunAishima:
various fixes necessary to get raster working to a reasonable extent. Note that rastering still has issues when changing exposure time for unknown reasons.

- unstaging detector correctly
- stop detector at end of collection
- send row index and protocol so only first row uses full zebra setup
- move detector configuration before arming
- use lsdc-server-2022-4.0 - contains most fixes to mxtools (but not all)

### DIFF
--- a/daq_macros.py
+++ b/daq_macros.py
@@ -3434,7 +3434,7 @@ def zebraDaqRasterBluesky(flyer, angle_start, num_images, scanWidth, imgWidth, e
                    x_end_um=x_vec_end, y_end_um=y_vec_end, z_end_um=z_vec_end, \
                    file_prefix=filePrefix, data_directory_name=data_directory_name,\
                    detector_dead_time=detectorDeadTime, scan_encoder=scanEncoder, change_state=changeState,\
-                   row_index=row_index, transmission=1)
+                   row_index=row_index, transmission=1, protocol="raster")
     yield from bp.fly([raster_flyer])
 
     logger.info("vector Done")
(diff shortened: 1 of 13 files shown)


daq_macros.py
```diff
@@ -1847,7 +1848,9 @@ def snakeRasterBluesky(rasterReqID, grain=""):
         xMotAbsoluteMove, xEnd, yMotAbsoluteMove, yEnd, zMotAbsoluteMove, zEnd = raster_positions(row, stepsize, omegaRad, rasterStartX, rasterStartY, rasterStartZ, row_index)
         vector = {'x': (xMotAbsoluteMove, xEnd), 'y': (yMotAbsoluteMove, yEnd), 'z': (zMotAbsoluteMove, zEnd)}
         yield from zebraDaqRasterBluesky(raster_flyer, omega, numsteps, img_width_per_cell * numsteps, img_width_per_cell, exptimePerCell, rasterFilePrefix,
-            data_directory_name, file_number_start, vector)
+            data_directory_name, file_number_start, row_index, vector)
+        raster_flyer.zebra.reset.put(1)  # reset after every row to make sure it is clear for the next row
+        time.sleep(0.2)
```
vshekar, line 1853: A comment about the reason for sleep would be helpful

JunAishima: it's in the comment for this commit, but it should be highlighted in the line. thanks for the reminder. will resolve when I add a comment

JunAishima: comment added in https://github.com/JunAishima/lsdc/commit/f61549d6351290cb52632e3be8aa3071c5f3db96

vshekar (approved):

PR #249, opened 2023-01-03
Update publish-docs.yml

vshekar:
Added graphviz to env to build docs based on existing marketplace action: https://github.com/marketplace/actions/setup-graphviz


JunAishima (approved):
looks good - I saw that this is the same configuration as used in the example at https://github.com/marketplace/actions/setup-graphviz

PR #250, opened 2023-01-03
Added Amx annealer code

RobertSchaffer1:
Code to unify AMX, FMX annealers

### DIFF
--- a/daq_macros.py
+++ b/daq_macros.py
@@ -29,7 +29,7 @@ from threading import Thread
 from config_params import *
 from kafka_producer import send_kafka_message
 
-from fmx_annealer import govStatusGet, govStateSet, annealer # for using annealer specific to FMX
+from fmx_annealer import govStatusGet, govStateSet, fmxAnnealer, amxAnnealer # for using annealer specific to FMX
 
 try:
   import ispybLib
@@ -3565,7 +3565,12 @@ def topViewCheckOn():
   setBlConfig(TOP_VIEW_CHECK,1)
 
 def anneal(annealTime=1.0):
-  if daq_utils.beamline == 'fmx':
+  if daq_utils.beamline == 'fmx' or daq_utils.beamline == 'amx':
+    if daq_utils.beamline == 'fmx':
+      annealer = fmxAnnealer
+    else:
+      annealer = amxAnnealer
+
     if not govStatusGet('SA'):
       daq_lib.gui_message('Not in Governor state SA, exiting')
       return -1
@@ -3582,25 +3587,20 @@ def anneal(annealTime=1.0):
 
     time.sleep(annealTime)
     annealer.air.put(0)
+    if daq_utils.beamline == 'fmx':
+      while not annealer.outStatus.get():
+        logger.info(f'anneal state after annealing: {annealer.outStatus.get()}')
+        time.sleep(0.1)
 
-    while not annealer.outStatus.get():
-      logger.info(f'anneal state after annealing: {annealer.outStatus.get()}')
-      time.sleep(0.1)
+    elif daq_utils.beamline == 'amx':
+      while annealer.inStatus.get():
+        logger.info(f'anneal state after annealing: {annealer.inStatus.get()}')
+        time.sleep(0.1)
 
     if govStateSet('SA') == -1:
       logger.error("not able to return to SA governor state")
       return -1
 
-  elif daq_utils.beamline == 'amx':
-    robotGovState = (getPvDesc("robotSaActive") or getPvDesc("humanSaActive"))
-    if (robotGovState):
-      setPvDesc("annealIn",1)
-      while (getPvDesc("annealStatus") != 1):
-        time.sleep(0.01)
-      time.sleep(float(annealTime))
-      setPvDesc("annealIn",0)
-    else:
-      daq_lib.gui_message("Anneal only in SA state!!")
   else:
     daq_lib.gui_message(f'Anneal not implemented for beamline {daq_utils.beamline}! Doing nothing')
   
--- a/fmx_annealer.py
+++ b/fmx_annealer.py
@@ -5,6 +5,7 @@ import epics
 import time
 
 import socket
+import daq_utils
 
 def blStrGet():
     """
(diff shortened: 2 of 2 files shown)


fmx_annealer.py
```diff
@@ -5,6 +5,7 @@
 import time
 
 import socket
```
JunAishima, line 7: no longer necessary now that you aren't checking hostname

daq_macros.py
```diff
@@ -3565,7 +3565,12 @@ def topViewCheckOn():
   setBlConfig(TOP_VIEW_CHECK,1)
 
 def anneal(annealTime=1.0):
-  if daq_utils.beamline == 'fmx':
+  if daq_utils.beamline == 'fmx' or daq_utils.beamline == 'amx':
```
JunAishima, line 3568: can you also add a TODO that all of this governor handling code can be removed to use standard code in LSDC2?

JunAishima (changes requested):
mostly minor things except the PV polling part

I foresee a big cleanup once we get LSDC2 as the main branch, though... this module will become a lot smaller, so the TODO is a good reminder for future us

daq_macros.py
```diff
@@ -29,7 +29,7 @@
 from config_params import *
 from kafka_producer import send_kafka_message
 
-from fmx_annealer import govStatusGet, govStateSet, annealer # for using annealer specific to FMX
+from fmx_annealer import govStatusGet, govStateSet, fmxAnnealer, amxAnnealer # for using annealer specific to FMX
```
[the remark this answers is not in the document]
RobertSchaffer1: Done

daq_macros.py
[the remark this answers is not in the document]
RobertSchaffer1: Makes sense, done

fmx_annealer.py
RobertSchaffer1: Agreed

daq_macros.py
RobertSchaffer1: Yup, done.

daq_macros.py
```diff
@@ -3582,25 +3587,20 @@ def anneal(annealTime=1.0):
 
     time.sleep(annealTime)
     annealer.air.put(0)
+    if daq_utils.beamline == 'fmx':
+      while not annealer.outStatus.get():
```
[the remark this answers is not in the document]
RobertSchaffer1: I'm not sure how much this improves code quality, but I've pushed a new version with this, let me know what you think.

fmx_annealer.py
```diff
@@ -106,6 +105,7 @@ def govStateSet(stateStr, configStr = 'Robot'):
     return
 
 from ophyd import PVPositioner, PVPositionerPC, Device, Component as Cpt, EpicsMotor, EpicsSignal, EpicsSignalRO
+from ophyd import Status
```
JunAishima, line 108: should be `SubscriptionStatus`

fmx_annealer.py
```diff
@@ -116,10 +116,52 @@ class FmxAnnealer(Device):
...
+
+
 class AmxAnnealer(Device):
     air = Cpt(EpicsSignal, '1}AnnealerAir-Sel')
     inStatus = Cpt(EpicsSignalRO, '2}AnnealerIn-Sts') # status: 0 (Not In), 1 (In)
 
+    def anneal(self, anneal_time):
+        def in_callback(value, old_value, **kwargs):
+            if old_value == 0 and value == 1:
+                logger.info(f'anneal state while annealing: {annealer.inStatus.get()}')
```
JunAishima, line 143: since we are already getting the value for `inStatus`, the message should just show `{value}` for the value

JunAishima (changes requested):

fmx_annealer.py
RobertSchaffer1: Nice catch, fixed

fmx_annealer.py
RobertSchaffer1: Good point.

JunAishima (approved):

fmx_annealer.py
RobertSchaffer1: Removed these completely, I don't think they add anything if it's just showing a number.

daq_macros.py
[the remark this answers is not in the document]
JunAishima: looked good enough - at least got rid of polling, though maybe not streamlined

fmx_annealer.py
JunAishima: by using logging, time would be shown, which gives an idea how long things really take (not just the requested anneal time). but fair enough to take it out completely

fmx_annealer.py
```diff
@@ -109,17 +105,64 @@ def govStateSet(stateStr, configStr = 'Robot'):
     return
 
 from ophyd import PVPositioner, PVPositionerPC, Device, Component as Cpt, EpicsMotor, EpicsSignal, EpicsSignalRO
```
RobertSchaffer1, line 107: @JunAishima  do you have any idea what happened here? My code doesn't work, I think because Cpt doesn't correspond to Component (possibly among other issues).

RobertSchaffer1: Or not, I'm very confused...

JunAishima: hmm, not sure. those names seem to be part of the main Ophyd namespace. what is the exception you get, or however you know it's not working?

RobertSchaffer1: Ah, nevermind, dumb mistake on my part wrt. the imports. My exception is the program complaining that it can't subscribe to inStatus, maybe this is because I haven't wrapped it in the Run Engine?

JunAishima: sounds likely - I haven't dug through the code to see what happens in that kind of case. you can probably do a bluesky.plan_stubs.wait instead of using the status's wait directly

RobertSchaffer1:
Tested at AMX, seems to work fine. I'm happy to merge, although we should test at FMX at the earliest convenience. I didn't test through the daq_macros (as governor is in M) only through the ophyd object.

RobertSchaffer1:
I merged this, I haven't been able to test at FMX due to the current configuration of the beamline, so we should test when we get the chance.

PR #253, opened 2023-01-13
Sb 166 always stop detector

JunAishima:
Use finalize_wrapper to create an updated plan that always stops the detector at the end of a data collection.

refs https://jira.nsls2.bnl.gov/browse/SB-166

### DIFF
--- a/daq_macros.py
+++ b/daq_macros.py
@@ -37,6 +37,7 @@ from scans import (zebra_daq_prep, setup_zebra_vector_scan,
                    setup_vector_program)
 import bluesky.plan_stubs as bps
 import bluesky.plans as bp
+from bluesky.preprocessors import finalize_wrapper
 
 try:
   import ispybLib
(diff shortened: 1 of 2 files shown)


daq_macros.py
```diff
@@ -1932,8 +1935,7 @@ def snakeRasterBluesky(rasterReqID, grain=""):
       db_lib.updateRequest(rasterRequest)
       daq_lib.set_field("xrecRasterFlag",rasterRequest["uid"])
       logger.info(f'setting xrecRasterFlag to: {rasterRequest["uid"]}')
-    raster_flyer.detector.cam.acquire.put(0)
-    raster_flyer.detector.super_unstage()
+    raster_flyer.detector.unstage()
```
RobertSchaffer1, line 1938: What does this function do?

RobertSchaffer1: Or I guess that's clear, but why are we changing it in for acquire.put(0)?

JunAishima: in this pull request, I removed hard-coding detector acquire stop into the main body of the collection code, which is replaced by the finalize function. `super_unstage()` was a previous attempt to just do `unstage()` of the detector instead of stopping the detector.

`unstage()` always needs to be done at the end of a collection after hardware is staged (in the case of this Eiger, setting up the prefix and data directory). the `stage()` function will do things like make sure the detector isn't staged already and get it ready to go. we are not using this to arm the detector, btw, because we want to use the time while arming to do other things (like governor changes). (though we are currently out of order with arming and governor changes.)

RobertSchaffer1 (approved):

PR #255, opened 2023-01-17
Added logic for case when detector is armed prior to scan

RobertSchaffer1:


RobertSchaffer1:
That looks a bit healthier...

daq_macros.py
```diff
@@ -1840,6 +1840,14 @@ def snakeRasterBluesky(rasterReqID, grain=""):
...
 
     raster_flyer.configure_detector(file_prefix=rasterFilePrefix, data_directory_name=data_directory_name)
+    if raster_flyer.detector.cam.armed.get() == 1:
+        daq_lib.gui_message('Detector is in armed state from previous collection! Stopping detector, but the user '
+                            'should check the most recent collection to determine if it was successful. Cancelling'
+                            'this collection, retry when ready.')
+        raster_flyer.detector.cam.acquire.put(0)
+        logger.warning("Detector was in the armed state prior to this attempted collection.")
+        return 0
+        #raise RuntimeError("Detector is armed, cannot continue in current state.")
```
JunAishima, line 1850: can you get rid of this last commented-out line?

JunAishima (changes requested):
please just remove that last line and otherwise looks good!

strictly speaking, stopping the detector won't be necessary because we are using the finalize_wrapper that will always stop it at the end of the function call, but it's good to see anyways.

daq_macros.py
RobertSchaffer1: Done.

JunAishima (approved):

PR #256, opened 2023-01-17
Added checks to prevent governor from making unsafe transitions

mskinner5278:
testing in progress


JunAishima (approved):
very useful to have these messages, especially for future debugging of governor and motor issues.

PR #259, opened 2023-01-23
Fix auto collect 2023 01

JunAishima:
Automated collection was not working correctly. This was due to zebraCamDaq() being run, and this called zebraDaqPrep(). I had tinkered with this function early in the conversion to Bluesky (and it wasn't working). restore old code for now


JunAishima:
closing as more changes were required (#259 )

PR #261, opened 2023-02-01
Remove ispyblib from gui

JunAishima:
remove ispybLib from GUI code to simplify dependencies and improve security of ISPyB by only accessing it from the server.


RobertSchaffer1 (approved):

JunAishima:
check this by running some data collections on AMX with the code merged in. dcid 60167695. I can see the field for crystal image and it looks correct - I can see that the file contains the heatmap as part of the crystal snapshot. merging

PR #262, opened 2023-02-01
Added a check to validate HDF5 result file

vshekar:
Validates whether an HDF5 result file can be opened by albula. 
If not a message will pop-up that the file has issues. Will most likely be removed once Albula2 is released

### DIFF
--- a/albulaUtils.py
+++ b/albulaUtils.py
@@ -89,16 +89,17 @@ def _albulaDispFile(filename):
         if not (currentMasterH5 == filename[0]):
             logger.info('reading file: %s' % filename[0])
             albulaSubFrame.loadFile(filename[0])
-            imgSeries = dectris.albula.DImageSeries(filename[0])
             currentMasterH5 = filename[0]
         logger.debug('reading image number %s' % filename[1])
-        albulaSubFrame.loadImage(imgSeries[filename[1]])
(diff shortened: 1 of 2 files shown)


albulaUtils.py
```diff
@@ -87,14 +90,31 @@ def _albulaDispFile(filename):
...
         imgSeries = dectris.albula.DimageSeries(filename[0])
         albulaSubFrame.loadImage(imgSeries[filename[1]])
         currentMasterH5 = filename[0]
-
+    except Exception as e:
+        logger.error(f'Albula exception: {e}')
+
+def validate_master_HDF5_file(filename):
+  """
+  Check if H5 file is valid
```
JunAishima, line 108: for now, validation consists of checking whether the data files exist and are readable. could you update this comment? I expect we could do more validation later on, so would like to keep the function name as it is

JunAishima (changes requested):

albulaUtils.py
vshekar: Added comment in 2ad40c0445b7ce7facec643c18bb611089f3d74e

JunAishima: looks good now. resolving

JunAishima (approved):

PR #264, opened 2023-02-02
Run fastdp gathering script fmx

JunAishima:
As requested by Wuxian, she noticed that the fast DP gathering script was not running, but was for LSDC1.


JunAishima:
re-implemented on the beamline today - works

vshekar (approved):

PR #268, opened 2023-02-17
Gonio will move for raster only under certain conditions

vshekar:
- State has to be SA
- Control has to be enabled
- Sample of selected collection is mounted


vshekar:
Solved issue described in https://jira.nsls2.bnl.gov/browse/SB-283

JunAishima (approved):

PR #271, opened 2023-02-25
Changes dewartree view:

vshekar:
- Added parameter table on hovering over qtreeview
- Clicking on will only show raster on screen and images on albula
- Right click menu to add parameters to form


vshekar:
Outdated code, no longer useful

PR #275, opened 2023-03-16
Fmx calc lifetime

vshekar:


raddoseLib.py
```diff
@@ -267,10 +268,10 @@ def fmx_expTime_to_10MGy(beamsizeV = 1.0, beamsizeH = 2.0,
                          verbose=verbose
                         )
     
-    logger.info("\n=== fmx_expTime_to_10MGy summary ===")
+    logger.info("\n=== fmx_expTime summary ===")
     dose1s = rd3d_out['DWD'].item()  # .item() to convert 1d array to scalar
     logger.info('Average Diffraction Weighted Dose for 1s exposure = {:f} MGy'.format(dose1s))
-    expTime10MGy = 10 / dose1s  # Experiment time to reach an average DWD of 10 MGy
+    expTime10MGy = avg_dwd / dose1s  # Experiment time to reach an average DWD of 10 MGy
```
JunAishima: shouldn't this comment and the one in the following line reflect the use of the dwd?

JunAishima: the variable name should also reflect that as well, instead of the hard-coded 10MGy name

JunAishima (changes requested):

JunAishima (approved):

PR #277, opened 2023-03-17
Fix multi col sb 315

JunAishima:


JunAishima:
I have made changes on FMX to make multiCol collection rows behave as rasters, by making them display the heatmap when selected, and omega to be moved. these changes will be put into this pull request before it is prepared for review.

vshekar (approved):

PR #282, opened 2023-04-05
Service user 2023 1

JunAishima:
- remove dzdo when killing existing LSDC servers - they should all be owned by the service user now
- check that a service user is running the server - stop if not. otherwise, calls to processing, writing to log files and other functions will be done by the human user, potentially leading to issues (such as requiring Kerberos for processing, SSH key issues...)

For now, we will discuss the issues of beamline staff having no access to the LSDC server terminal, which may be problematic when handling operational issues. Will leave this issue draft until that is resolved.

### DIFF
--- a/daq_main2.py
+++ b/daq_main2.py
@@ -12,6 +12,7 @@ from daq_lib import *
 from robot_lib import *
 from beamline_lib import *
 from gov_lib import setGovRobot
+import getpass
 
 import logging
 from logging import handlers
@@ -27,6 +28,13 @@ handler2.setFormatter(myformat)
 logger.addHandler(handler1)
 logger.addHandler(handler2)
 
+if not getpass.getuser().startswith("lsdc-"):
+    message = "LSDC server not being started by a LSDC service user account, aborting!"
+    print(message)
+    logger.error(message)
+    sys.exit(1)
+else:
+    print(f"continuing as we are using a service user: {getpass.getuser()}")
 sitefilename = ""
 global command_list,immediate_command_list,z
 command_list = []
--- a/daq_main2.py
+++ b/daq_main2.py
@@ -5,6 +5,7 @@ The main server for the LSDC system
 import sys
 import os
 from daq_main_common import pybass_init, run_server
+from config_params import LSDC_SERVICE_USERS
 
 #TODO understand why imports are required here - GUI requires imports in daq_main_common
 from daq_macros import *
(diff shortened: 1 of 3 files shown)


lsdcServer
```diff
@@ -1,3 +1,3 @@
-dzdo pkill -KILL -f daq_main2.py
+pkill -KILL -f daq_main2.py
```
danielballan, line 1: Have you considered using a pidfile? Relying on `[p]kill` in this way is unusual.

daq_main2.py
```diff
@@ -27,6 +28,13 @@
 logger.addHandler(handler1)
 logger.addHandler(handler2)
 
+if not getpass.getuser().startswith("lsdc-"):
```
danielballan, line 31: Is the set of allowed accounts very large? Could they be spelled out fully and checked for an exact match, to future-proof against possible other lsdc-related accounts being accidentally used here?

JunAishima: there are currently only 3 "lsdc-" service accounts. I did not want to hard-code these into the code but would it be more satisfactory to put them in here? I will need to think about how best to do this.

lsdcServer
JunAishima: This was a holdover from manually killing all other instances of the LSDC server - they could have been started by multiple local contacts.
Thinking about this a bit more, would I be able to count on systemd killing all processes spawned by launching LSDC server? if so, then I don't even need the line at all.
I don't know enough about how the main process and related processes are handled by systemd

daq_main2.py
danielballan: I think that's safer. We add beamlines rarely, and this will avoid potential unintended permissiveness.

lsdcServer
danielballan: OK, we can let this through for now but it should be revisited.

daq_main2.py
JunAishima: fixed in https://github.com/NSLS-II/lsdc/pull/282/commits/be9ddd08337e4be053ee7ff6b8b012b0e4dfe77d . if you are happy with this fix, please resolve!

lsdcServer
JunAishima: actually, this issue seems to be resolved with my Ansible pull request, where using Type=exec and having Conflict settings on the unit ensure that the "other" of lsdc-server and lsdc-server-dev that is not currently running will be stopped, if it had been running.

if this is satisfactory, please resolve this conversation

tacaswell (approved):

PR #286, opened 2023-04-25
Split GUI code

vshekar:
Moved LSDC GUI code from one file to a gui module 
Fixed imports according to what is required by classes
Formatted GUI code and sorted import
Started LSDC at FMX explored existing collections


gui/dialog/puck_dialog.py
```diff
@@ -0,0 +1,94 @@
+import logging
+import typing
+from typing import Optional
```
JunAishima, line 3: is there any reason this import needs to be here? seems like the Optional item is never used

gui/control_main.py
```diff
@@ -4852,7 +4852,7 @@ def initUI(self):
         self.immediate_comm_pv = PV(daq_utils.beamlineComm + "immediate_command_s")
         self.stillModeStatePV = PV(daq_utils.pvLookupDict["stillModeStatus"])
         self.progressDialog = QtWidgets.QProgressDialog()
-        self.progressDialog.setCancelButtonText(QString("Cancel"))
+        self.progressDialog.setCancelButtonText("Cancel")
```
JunAishima, line 4850: according to the documentation, it seems to be ok to use regular text. was there any reason why QString was used before?

JunAishima: I have seen this in at least one other place, perhaps requirements changed in Qt or PyQt some time ago...

gui/dialog/puck_dialog.py
vshekar: Removed those imports in https://github.com/NSLS-II/lsdc/pull/286/commits/61f92ac3c9f28c816fcf64848f1240f421b9a8a0

gui/control_main.py
JunAishima: you provided the explanation that this might have been an old requirement for the buttons and what it was being covered up with "QString = str"... resolving.

JunAishima (approved):
a couple of good comments:

- linting is nice, will make it easier to work with
- liked the imports in dialog - looks like it makes the imports more intuitive

PR #289, opened 2023-05-02
Synchronize arming

RobertSchaffer1:
@JunAishima @vshekar Changes to parallelize the arming and governor moves, requires https://github.com/NSLS-II-AMX/mxtools/pull/33

### DIFF
--- a/daq_macros.py
+++ b/daq_macros.py
@@ -1857,11 +1857,13 @@ def snakeRasterBluesky(rasterReqID, grain=""):
                             'this collection, retry when ready.')
         logger.warning("Detector was in the armed state prior to this attempted collection.")
         return 0
+    start_time = time.time()
     arm_status = raster_flyer.detector_arm(angle_start=omega, img_width=img_width_per_cell, total_num_images=totalImages, exposure_period_per_image=exptimePerCell, file_prefix=rasterFilePrefix,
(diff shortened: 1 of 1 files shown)


daq_macros.py
```diff
@@ -1854,17 +1851,23 @@ def snakeRasterBluesky(rasterReqID, grain=""):
...
+    arm_status = raster_flyer.detector_arm(angle_start=omega, img_width=img_width_per_cell, total_num_images=totalImages, exposure_period_per_image=exptimePerCell, file_prefix=rasterFilePrefix,
                        data_directory_name=data_directory_name, file_number_start=file_number_start, x_beam=xbeam, y_beam=ybeam, wavelength=wave, det_distance_m=detDist,
                        num_images_per_file=numsteps)
+    govStatus = gov_lib.setGovRobot(gov_robot, "DA")
+    if govStatus.exception():
+      logger.error(f"Problem during start-of-raster governor move, aborting! exception: {govStatus.exception()}")
+      arm_status.wait()
+      raster_flyer.detector.cam.acquire.put(0)
+      return
+    arm_status.wait()
```
JunAishima: every time we have an arm_status.wait(), can we output the time difference from when arming is started?

JunAishima (changes requested):
logging for arm+governor time would be very nice to have here

daq_macros.py
RobertSchaffer1: Sure, I suppose.

RobertSchaffer1: Done.

JunAishima (approved):

PR #291, opened 2023-05-04
Remove light buttons

vshekar:


gui/control_main.py
```diff
@@ -3718,8 +3720,10 @@ def addRequestsToAllSelectedCB(self):
                 elif itemDataType == "request":
                     selectedSampleRequest = db_lib.getRequestByID(item.data(32))
                     self.selectedSampleID = selectedSampleRequest["sample"]
-                
-                if self.selectedSampleID in samplesConsidered: # If a request is already added to the sample, move on
+
```
JunAishima, line 3723: can you remove this code change? I know it looks like an obvious linting fix, but shouldn't be here

JunAishima (changes requested):

JunAishima:
do we have an idea of the timeline for when the light controls would be returned? if it's going to be returned during the cycle, we should keep the code as-is and make temporary changes in the "production" branch

vshekar:
I think you're right it should be returned in the next cycle or two, so it might be better to disable in the production branch rather than on github. @RobertSchaffer1 opinions?

gui/control_main.py
vshekar: Removed in 7a3c09728df0e70791374f409280d708f8fcc6e5

JunAishima (approved):

PR #295, opened 2023-05-19
Lifetime fixes

vshekar:
Fixed a case if osc_end_ledit text box is empty, the GUI crashes
Changed the way raddose config file is edited. Instead of opening and closing each line after every edit, the file is opened and closed only once and edits are made at the same time
Added process_id to raddose output to differentiate between users calculating at the same time

### DIFF
--- a/raddoseLib.py
+++ b/raddoseLib.py
@@ -8,53 +8,80 @@ import sys
 import os.path
 import os
 import logging
+
 logger = logging.getLogger(__name__)
 
-#12/19 - See Martin about this code!
+# 12/19 - See Martin about this code!
+
 
 def replaceLines(filename, replacementStrings):
     lines = []
-    with open(filename, 'r') as f:
+    with open(filename, "r") as f:
         for line in f.readlines():
             for text_to_search, replacement_text in replacementStrings.items():
                 if text_to_search in line:
                     line = replacement_text
                     break
             lines.append(line)
-    
-    with open(filename, 'w') as f:
+
+    with open(filename, "w") as f:
         f.writelines(lines)
 
-    
+
 def run_rd3d(inputFileName):
-    prc = subprocess.Popen(["java", "-jar", os.environ['CONFIGDIR'] + "/raddose3d.jar", "-i", inputFileName, "-p", f"rd3d/rd3d_{os.getpid()}_"],
+    prc = subprocess.Popen(
+        [
+            "java",
+            "-jar",
+            os.environ["CONFIGDIR"] + "/raddose3d.jar",
+            "-i",
+            inputFileName,
+            "-p",
+            f"rd3d/rd3d_{os.getpid()}_",
+        ],
         stdout=subprocess.PIPE,
-        universal_newlines=True)
+        universal_newlines=True,
+    )
     out = prc.communicate()[0]
     return out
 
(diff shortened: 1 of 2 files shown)


raddoseLib.py
```diff
@@ -8,6 +8,7 @@
 import os.path
 import os
 import logging
+import traceback
```
JunAishima: is there a particular reason you are putting in the traceback here?

raddoseLib.py
```diff
@@ -19,8 +20,22 @@ def replaceLine(file,searchExp,replaceExp):
...
+                    break
+            lines.append(line)
+    
+    with open(filename, 'w') as f:
+        f.writelines(lines)
+
+    
 def run_rd3d(inputFileName):
-    prc = subprocess.Popen(["java", "-jar", os.environ['CONFIGDIR'] + "/raddose3d.jar", "-i", inputFileName, "-p", "rd3d/rd3d_"],
+    prc = subprocess.Popen(["java", "-jar", os.environ['CONFIGDIR'] + "/raddose3d.jar", "-i", inputFileName, "-p", f"rd3d/rd3d_{os.getpid()}_"],
```
JunAishima, line 30: would it make sense to make the raddose3d.jar file an RPM and deploy? It doesn't seem to belong in the CONFIGDIR or LSDC.

raddoseLib.py
```diff
@@ -98,31 +113,35 @@ def rd3d_calc(flux=3.5e12, energy=12.66,
...
-    replaceLine(inputFilePath,"TRANSLATEPERDEGREE",
-                'TRANSLATEPERDEGREE {:0.4f} {:0.4f} {:0.4f}\n'.format(translatePerDegX,translatePerDegY,translatePerDegZ))
-    replaceLine(inputFilePath,"DIMENSION",'DIMENSION {:0.1f} {:0.1f} {:0.1f}\n'.format(dimX,dimY,dimZ))
-    replaceLine(inputFilePath,"PIXELSPERMICRON",'PIXELSPERMICRON {:0.1f}\n'.format(pixelsPerMicron))
-    replaceLine(inputFilePath,"ANGULARRESOLUTION",'ANGULARRESOLUTION {:0.1f}\n'.format(angularResolution))
-    replaceLine(inputFilePath,"STARTOFFSET",
-                'STARTOFFSET {:f} {:f} {:f}\n'.format(startOffsetX,startOffsetY,startOffsetZ))    
-    
-    out = run_rd3d(inputFilePath)
+
```
JunAishima, line 116: I wonder whether this file is templated heavily enough that we should just use jinja2?

JunAishima (changes requested):

raddoseLib.py
vshekar: I would say so, currently whenever LSDC starts up, it checks if the working dir has a raddose folder, if it doesn't it copies files from the bnlpx_config to the working dir and runs raddose using those files. Probably a historical way of making it run and still being used

JunAishima: perhaps we should re-think a lot of what is being done here, but I have proposed a possible solution or two for the raddose RPM issue - https://jira.nsls2.bnl.gov/browse/SB-447

JunAishima (approved):

PR #296, opened 2023-05-25
Fixed wrong parameters passed to mount function when retrying after error

vshekar:
Would crash inside mount saying puckPos and pinPos could not be added because one of them is a string. That's because it was trying to add pinPos and sampID


JunAishima (approved):
aah, looks like you noticed a kwargs passing issue, too

JunAishima (approved):

JunAishima:
for @danielballan and @tacaswell - non-critical fix. lowest priority.

Data security implications: none

PR #297, opened 2023-05-26
Initial attempt at extracting certain frames from the main window

vshekar:
This is a first attempt at removing the definition of certain parts of the acquisition frame and moving it to their own files. 
This kind of modularization will allow another beamline to inherit, pick and choose what it wants to modify by class.

This pull request consists of extracting the following frames
- Vector Params
- MultiCol params
- Characterize params
- Raster params
- Processing options params

### DIFF
--- a/gui/acquisition/__init__.py
+++ b/gui/acquisition/__init__.py
@@ -1,3 +1,3 @@
 from .vector_params import VectorParamsFrame
 from .multicol_params import MultiColParamsFrame
-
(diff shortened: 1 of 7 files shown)


JunAishima:
do you mean that they would modify certain classes in a particular file instead of modifying control_main as a whole?
this brings up the issue of how forking LSDC affects deployment. if we do end up packaging LSDC into a pypi package -> conda package, a fork would have to have an entirely different pipeline, from pypi to conda package to conda environment.

PR #299, opened 2023-06-06
Centralize collection directory

JunAishima:
- Changes to ensure server and GUI are writing to the same directory
- GUI
  - Not allowed to start if the directory where it is started is not the same as where the server is started
  - Will be terminated if the visit directory is changed while it is running
- Server
  - Not allowed to start if the `current_visit_lsdc_<tla>` file does not exist or the `CURRENT_VISIT_DIR` variable does not point to a proposal directory-like name

### DIFF
--- a/gui/control_main.py
+++ b/gui/control_main.py
@@ -1378,21 +1378,21 @@ class ControlMain(QtWidgets.QMainWindow):
             self.vidActionRasterDefRadio.setDisabled(True)
             self.vidActionDefineCenterRadio.setDisabled(True)
 
-        hutchCornerCamThread = VideoThread(
+        self.hutchCornerCamThread = VideoThread(
             parent=self, delay=HUTCH_TIMER_DELAY, url=getBlConfig("hutchCornerCamURL")
         )
-        hutchCornerCamThread.frame_ready.connect(
+        self.hutchCornerCamThread.frame_ready.connect(
             lambda frame: self.updateCam(self.pixmap_item_HutchCorner, frame)
         )
-        hutchCornerCamThread.start()
+        self.hutchCornerCamThread.start()
 
-        hutchTopCamThread = VideoThread(
+        self.hutchTopCamThread = VideoThread(
             parent=self, delay=HUTCH_TIMER_DELAY, url=getBlConfig("hutchTopCamURL")
         )
-        hutchTopCamThread.frame_ready.connect(
+        self.hutchTopCamThread.frame_ready.connect(
             lambda frame: self.updateCam(self.pixmap_item_HutchTop, frame)
         )
-        hutchTopCamThread.start()
+        self.hutchTopCamThread.start()
         serverCheckThread = ServerCheckThread(
             parent=self, delay=SERVER_CHECK_DELAY)
         serverCheckThread.visit_dir_changed.connect(QApplication.instance().quit)
@@ -4958,7 +4958,11 @@ class ControlMain(QtWidgets.QMainWindow):
             self.popupServerMessage("You don't have control")
 
     def closeAll(self):
-        QtWidgets.QApplication.closeAllWindows()
+        self.hutchCornerCamThread.stop()
+        self.hutchTopCamThread.stop()
+        self.hutchCornerCamThread.wait()
+        self.hutchTopCamThread.wait()
+        QtWidgets.QApplication.instance().quit()
 
     def initCallbacks(self):
         self.beamSizeSignal.connect(self.processBeamSize)
--- a/threads.py
+++ b/threads.py
@@ -56,15 +56,19 @@ class VideoThread(QThread):
         self.url = url
         self.camera_object = camera_object
         self.showing_error = False
+        self.is_running = True
         QThread.__init__(self, *args, **kwargs)
     
     def updateCam(self, camera_object):
         self.camera_object = camera_object
         
     def run(self):
-        while True:
+        while self.is_running:
             self.camera_refresh()
             self.msleep(self.delay)
+    
+    def stop(self):
+        self.is_running = False
 
 
 class RaddoseThread(QThread):


utils/healthcheck.py
```diff
@@ -105,20 +105,43 @@ def handle_fail(check):
...
-        check_curr_visit_dir.remediation = "CURRENT_VISIT_DIR is empty"
+    if os.environ[visit_dir_env_var] == "":
+        check_curr_visit_dir.remediation = f"{visit_dir_env_var} is empty"
         return False
-    current_visit_dir = Path(os.environ["CURRENT_VISIT_DIR"])
+    current_visit_dir = Path(os.environ[visit_dir_env_var])
     if not current_visit_dir.exists():
         check_curr_visit_dir.remediation = (
-            f"CURRENT_VISIT_DIR = {current_visit_dir} does not exist"
+            f"{visit_dir_env_var} = {current_visit_dir} does not exist"
```
JunAishima, line 121: can you also include a check of whether the visit directory is writeable to the user who runs the server?
https://stackoverflow.com/questions/2113427/determining-whether-a-directory-is-writeable
Two good candidates for methods to do this are above: the access method, and "zak"'s method of writing a file and handling an exception if necessary

vshekar: Added in https://github.com/NSLS-II/lsdc/pull/299/commits/88b4ede2c957339cd59809ddb79b6a45199e2665

JunAishima:
@vshekar - some information about the branches of LSDC in the current condition with the fixes from this pull request vs. what was running before

current branches containing the fixes:
AMX: 2023-2-amxfmx-production-dev
FMX: 2023-2-amxfmx-production-dev

previous branches that we must roll back to if necessary (pre-fix branches):
AMX: 2023-2-amxfmx-production (also archive-2023-05-16)
FMX: 2023-2-amxfmx-production-transmission-fixes

daq_main2.py
```diff
@@ -35,6 +35,8 @@
 logger.addHandler(handler1)
 logger.addHandler(handler2)
 
+perform_server_checks()
+setBlConfig("visitDirectory", os.getcwd())
```
tacaswell, line 39: It is a bit opaque to me that after `perform_sever_checks` that we are sure that the cwd is a valid visit directory.  It might be better to have a helper function that looks like

```python
def get_visit_directory():
    candidate = os.getcwd()
    if is_valid(candidate):
        raise ...
    return candidate

```

I think this would better ensure that the check does not get accidentally lost and reduces the number of places `os.getcwd()` exists in the code.

top_view.py
```diff
@@ -31,10 +31,10 @@ def wait90TopviewThread(gov_robot, prefix1,prefix90):
...
     topViewWrite()
-    topViewSnap(prefix90,os.getcwd()+"/pinAlign",1)
+    topViewSnap(prefix90,getBlConfig("visitDirectory")+"/pinAlign",1)
     snapshot1Name = prefix1+"_001.jpg"
     snapshot2Name = prefix90+"_001.jpg"
-    if (not filecmp.cmp(os.getcwd()+"/pinAlign/"+snapshot1Name,os.getcwd()+"/pinAlign/"+snapshot2Name)): #this would mean something is wrong if true because the pictures are identical
+    if (not filecmp.cmp(getBlConfig("visitDirectory")+"/pinAlign/"+snapshot1Name,getBlConfig("visitDirectory")+"/pinAlign/"+snapshot2Name)): #this would mean something is wrong if true because the pictures are identical
       comm_s = os.environ["LSDCHOME"] + "/runPinAlign.py " + snapshot1Name + " " + snapshot2Name
       logger.info(comm_s)
       lines = os.popen(comm_s).readlines()
```
tacaswell, line 40:
```suggestion
      lines = os.popen(comm_s, cwd=.getBlConfig("visitDirectory"))readlines()
```

will make sure that what ever subprocess you are working with will run in the visit directory as the cwd.

gui/data_loc_info.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, parent: "ControlMain"):
         self.hBoxDPathParams1 = QtWidgets.QHBoxLayout()
         self.basePathLabel = QtWidgets.QLabel("Base Path:")
         self.base_path_ledit = QtWidgets.QLabel()  # leave editable for now
```
tacaswell, line 25: This is editable but does nothing?

gui/control_main.py
```diff
@@ -1392,6 +1393,10 @@ def createSampleTab(self):
             lambda frame: self.updateCam(self.pixmap_item_HutchTop, frame)
         )
         hutchTopCamThread.start()
+        serverCheckThread = ServerCheckThread(
+            parent=self, delay=SERVER_CHECK_DELAY)
+        serverCheckThread.visit_dir_changed.connect(self.close)
```
tacaswell, line 1398: is `app.exit` (or similar) a better choice here? It does look like `def closeEvent` will exit, but directly exiting the app here may be more reliable?

tacaswell (approved):
I left a few comments that I think would reduce edge cases, but as-is this PR achieves the goals of:

 - only starting the server if it is in a directory with the correct structure
 - the GUI exits with in 2 seconds of the server directory being different than the cwd

daq_main2.py
JunAishima: I think I see what you are getting at with the other Slack conversation - as the visit directory gets changed as the server is kept running for multiple user visits, how can we ensure that it is always pointing to a good visit directory?

the medium term plan is to separate LSDC server from its cwd so that it will always write to the visit directory. the other useful (to local contact) output from the way the server is run is the log file. we will plan on using an environment variable to set where the logging output will be.

tacaswell: This is the only place in the codebase that `setBlConfig("visitDirectory", ...)` is called, how can it be changed?

I was more worried that at some point in the future `perform_server_checks` would stop checking `os.getcwd()` and we would lose control of it again.  A function that does the validation and return the valid path (or raises) would be more robust (and more testable!) it the medium term.  It would also easily allow for validating setting this later.

gui/data_loc_info.py
JunAishima: the comment is misleading as QLabel() is not editable. fixed in https://github.com/NSLS-II/lsdc/pull/299/commits/d785be05470ca0d74e3bc46c144c7d40952727b7

daq_main2.py
JunAishima: as part of the changes to separate the LSDC server from cwd as mentioned above, we could implement this suggestion when we make the server persist across multiple visits, at which point the check would become important as a server check won't necessarily be performed between different visits.

JunAishima:
Before: started at commit 02aca1 of LSDC, on 2023-2-amxfmx-production-dev, which includes up to commit 88b4ed of the centralize-collection-directory branch.

output with self.close():
` The server visit directory has changed, stopping!`

After: merged in latest centralize-collection-directory into 2023-2-amxfmx-production-dev. this includes the QApplication fix, in addition to top view fix for cwd. The latter was not checked.
output with QApplication.getInstance().quit:
```
The server visit directory has changed, stopping!
QPixmap::fromImage: QPixmap cannot be created without a QGuiApplication
QPixmap: Must construct a QGuiApplication before a QPixmap
```

tacaswell:
> QPixmap::fromImage: QPixmap cannot be created without a QGuiApplication
> QPixmap: Must construct a QGuiApplication before a QPixmap

That suggests that you have a slot someplace that does not check if the application is doing down before it does some work

vshekar:
That is most likely from a QThread that is getting camera images in the background to put in the UI. I can add a fix to clean up these threads before the quit is called either in this PR or a separate one

JunAishima:
I'm still getting `QPixmap: Must construct a QGuiApplication before a QPixmap` error as the GUI is being killed. this is slightly different from before in that the first line of the previous error is no longer shown.

PR #300, opened 2023-06-15
Added code to save FMX flux reference after setE

vshekar:
- Also moves to correct governor states after setE

### DIFF
--- a/setenergy_lsdc.py
+++ b/setenergy_lsdc.py
@@ -152,6 +152,10 @@ slits1 = Slits('XF:17IDA-OP:FMX{Slt:1', name='slits1', labels=['fmx'])
 ## Light
 light = YMotor('XF:17IDC-ES:FMX{Light:1', name='lightY')
 
+## Temporary: Disable motor alarm until end switch is fixed
+from ophyd.utils.epics_pvs import AlarmSeverity
+light.y.tolerated_alarm = AlarmSeverity.MAJOR
+
 ## Goniometer Stack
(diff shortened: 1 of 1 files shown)


setenergy_lsdc.py
```diff
@@ -1002,3 +1008,225 @@ def beam_center_align(transSet='All'):
...
+    flux_df.at[slit1Gap, 'Keithley current [A]'] = keithley.get()
+    flux_df.at[slit1Gap, 'Keithley flux [ph/s]'] = get_fluxKeithley()
+    flux_df.at[slit1Gap, 'BPM1 sum [A]'] = bpm1.sum_all.get()
+    # TEMP FIX: flux_df.at[slit1Gap, 'BPM4 sum [A]'] = bpm4.sum_all.get()
+    flux_df.at[slit1Gap, 'BPM4 sum [A]'] = 0
+
+def fmx_flux_reference(slit1GapList = [2000, 1000, 600, 400], slit1GapDefault = 1000, transSet='All'):
+    """
+    Sets Slit 1 X gap and Slit 1 Y gap to a list of settings,
+    and returns flux reference values in a pandas DataFrame.
```
JunAishima, line 1074: this line suggests that the values are returned to the user/caller - but it's actually just storing the global for future use.

JunAishima (changes requested):

setenergy_lsdc.py
vshekar: Fixed in https://github.com/NSLS-II/lsdc/pull/300/commits/323d1c837e0a6805b249d5d587d4194e90727675

JunAishima (approved):

PR #301, opened 2023-06-26
all production server and GUI instances will be run from lsdc_<tla>_prod

JunAishima:
* development clones of LSDC will be located at /nsls2/software/mx/daq/lsdc_<tla> (original clones located in the original location), while production clones have been checked out to /nsls2/software/mx/daq/lsdc_<tla>_prod

### DIFF
--- a/bin/lsdcServer_amx.cmd
+++ b/bin/lsdcServer_amx.cmd
@@ -1,7 +1,7 @@
 #!/bin/bash -l
 export PROJDIR=/nsls2/software/mx/daq/
 export CONFIGDIR=${PROJDIR}bnlpx_config/
-export LSDCHOME=${PROJDIR}lsdc_amx
+export LSDCHOME=${PROJDIR}lsdc_amx_prod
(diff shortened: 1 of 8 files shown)


bin/lsdcServer_amx.cmd
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash -l
 export PROJDIR=/nsls2/software/mx/daq/
 export CONFIGDIR=${PROJDIR}bnlpx_config/
-export LSDCHOME=${PROJDIR}lsdc_amx
```
tacaswell, line 4: Is the `${CONFIGDIR}daq_env_amx.txt` controlled by us or the user?  If it is the later anything we are going to rely on being what we set it to should probably be set _after_ the bl specific file.

Similarly, activating a conda env can set ENVS so maybe do that first?

JunAishima: the entire ${CONFIGDIR} should not be editable by the user, we have full control of it.

I have found that it can be tricky to set ENVs, so would rather not touch them if this works

tacaswell (approved):

PR #304, opened 2023-06-29
use subprocess.Popen which has the cwd kwarg

JunAishima:
* os.popen does not have the cwd arg, and this was preventing top_view from working correctly

### DIFF
--- a/top_view.py
+++ b/top_view.py
@@ -36,10 +36,10 @@ def wait90TopviewThread(gov_robot, prefix1,prefix90):
     snapshot1Name = prefix1+"_001.jpg"
     snapshot2Name = prefix90+"_001.jpg"
     if (not filecmp.cmp(getBlConfig("visitDirectory")+"/pinAlign/"+snapshot1Name,getBlConfig("visitDirectory")+"/pinAlign/"+snapshot2Name)): #this would mean something is wrong if true because the pictures are identical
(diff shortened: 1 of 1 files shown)


top_view.py
```diff
@@ -37,7 +38,8 @@ def wait90TopviewThread(gov_robot, prefix1,prefix90):
     if (not filecmp.cmp(getBlConfig("visitDirectory")+"/pinAlign/"+snapshot1Name,getBlConfig("visitDirectory")+"/pinAlign/"+snapshot2Name)): #this would mean something is wrong if true because the pictures are identical
       comm_s = os.environ["LSDCHOME"] + "/runPinAlign.py " + snapshot1Name + " " + snapshot2Name
       logger.info(comm_s)
-      lines = os.popen(comm_s, cwd=getBlConfig("visitDirectory")).readlines()
+      process = subprocess.Popen(comm_s, cwd=getBlConfig("visitDirectory"), stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
tacaswell, line 41: Sorry about that!

As long we are moving over to `subprocess` also switch to `run()`?

https://docs.python.org/3/library/subprocess.html#subprocess.run

vshekar: Fixed in https://github.com/NSLS-II/lsdc/pull/304/commits/4f424d88a96ca180c8ff6a37a97c6c89aaadfaf7

tacaswell (approved):

PR #306, opened 2023-06-30
Compliance

mskinner5278:
These are the changes that were required to meet compliance requirements.


tacaswell:
In its current state I do not think this PR in re-viewable.  The diff in +7k/-6k LOC which on a quick skim appear to be mostly new features.  This also appears to have many commits that were previously merged to the master branch and several commits with `[DONOTMERGE]` in their commit message.

This either needs to have the git history cleaned up (maybe merge all of the non-compliance changes to the target branch first?) or be accompanied by significant narrative explaining what all of these changes are.

tacaswell (changes requested):
This PR is not in a reviewable state.

PR #312, opened 2023-07-10
Adding Zoom 5 (or Zoom N)

vshekar:
Pull request that replaces the zoom radio buttons with a zoom slider. The zoom slider is not static and can be modified based on beamline config parameters

### DIFF
--- a/daq_macros.py
+++ b/daq_macros.py
@@ -86,131 +86,84 @@ def abortBS():
     except super_state_machine.errors.TransitionError:
       logger.error("caught BS")
 
-def changeImageCenterLowMag(x,y,czoom):
-  zoom = int(czoom)
-  zoomMinXRBV = getPvDesc("lowMagZoomMinXRBV")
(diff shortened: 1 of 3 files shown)


gui/camera/scene.py
```diff
@@ -0,0 +1,59 @@
...
+from qtpy.QtWidgets import QGraphicsView
+from qtpy.QtCore import Signal, Qt
+import daq_utils
+import typing
+
+if typing.TYPE_CHECKING:
+    from gui.control_main import ControlMain
+
+
+class CustomView(QGraphicsView):
```
vshekar, line 11: Created CustomView mainly to handle events that happen when the sample cam is in focus. It handles the following events:
- Using the mouse wheel (connects to `handle_wheel` method in `gui.camera.zoom_widget.ZoomSlider`
- Detects whether Ctrl key is pressed and the mouse is clicked on the sample cam
- Triggers center_on_click when all conditions above are met

PR #318, opened 2023-07-21
Removed hard coded return statements to properly return robot failures

vshekar:
Robot mount and unmount did not handle failures, fixed now.

### DIFF
--- a/robot_lib.py
+++ b/robot_lib.py
@@ -33,7 +33,6 @@ def unmountRobotSample(gov_robot, puck_pos, pin_pos, samp_id):
   if status != UNMOUNT_STEP_SUCCESSFUL:
       return status
   
-  status = UNMOUNT_SUCCESSFUL
   if robot.control_type() == "Bluesky":
     RE(robot.unmount(gov_robot, puck_pos, pin_pos, samp_id))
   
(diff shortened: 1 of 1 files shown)


robot_lib.py
```diff
@@ -26,21 +26,22 @@ def mountRobotSample(gov_robot, puck_pos, pin_pos, samp_id, **kwargs):
...
   status = robot.postMount(gov_robot, puck_pos, pin_pos, samp_id)
-  return MOUNT_SUCCESSFUL  # TODO hard-coded for testing
+  return status
 
 def unmountRobotSample(gov_robot, puck_pos, pin_pos, samp_id):
   status = robot.preUnmount(gov_robot, puck_pos, pin_pos, samp_id)
   if status != UNMOUNT_STEP_SUCCESSFUL:
       return status
+  
+  status = UNMOUNT_SUCCESSFUL
```
JunAishima, line 36: are these necessary? status is defined for both cases below, it shouldn't be necessary to be defined before then

vshekar: You're right fixed in https://github.com/NSLS-II/lsdc/pull/318/commits/64425a2ff92288aa04b06134298f0e168acaf99d

danielballan (approved):

PR #320, opened 2023-07-27
Fixed missing detector arming in standard collection

mskinner5278:


daq_macros.py
```diff
@@ -3433,10 +3434,21 @@ def zebraDaqBluesky(flyer, angle_start, num_images, scanWidth, imgWidth, exposur
                            'change_state':changeState, 'transmission':vector_params["transmission"],
                            'data_path':data_path}
     start_time = time.time()
-    #flyer.detector_arm(**required_parameters)
+    flyer.detector_arm(**required_parameters)
```
JunAishima, line 3437: this is not correct. look at the flyer.detector_arm() code and see what its return value is and how your code needs to be changed.

JunAishima (changes requested):
fix how to use flyer.detector_arm()

daq_macros.py
mskinner5278: We were getting NoneType back instead of the status object, so I have moved the creation of the status object to the collection to allow operations to continue.

PR #321, opened 2023-08-01
Qtsignals refactor

vshekar:
Previously, any Epics PV that triggers a change in the QT UI would do the following:
  - Define EPICS PV
  - Define QtSignal
  - EPICS callback that triggers the QT signal
  - QT Signal then runs a callback that does some work in the frontend
In practically every case, the EPICS callback simply takes the PV value and passes it to the qt signal. This leads to repeating code and very little variation (not DRY)
`EpicsQObject` attempts to address that. There are escape hatches in case there is a need for custom epics callbacks

### DIFF
--- a/gui/epics_signal.py
+++ b/gui/epics_signal.py
@@ -1,6 +1,6 @@
 from qtpy.QtCore import Signal, QObject
 from epics import PV
-
+from typing import Optional, Type
 
 class EpicsQObject(QObject):
     """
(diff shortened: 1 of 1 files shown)


gui/epics_signal.py
```diff
@@ -0,0 +1,40 @@
...
+    epics_pv_changed = Signal(object)
+
+    def __init__(self, pv_name, qt_callback, use_string=False, pv_callback=None, custom_pv=None):
+        super().__init__()
+        self.use_string = use_string
+        # Define the pyepics PV and its callback
+        if not pv_callback:
+            pv_callback = self.on_pv_changed
+
+        if custom_pv:
```
JunAishima: There are two things strange here:
* Is this a PV object or a function? it looks like a function because `self.pv` calls it like that
* I see no instance of `EpicsQObject` that uses this kwarg or even as a positional arg - is this necessary?

JunAishima (commented):
Removing boilerplate and using f-strings instead of raw addition are very welcome, but it is unclear whether the `custom_pv` kwarg is necessary with the current LSDC code and this PR.

PR #329, opened 2023-09-19
Puck visibility

vshekar:
This PR adds code to the dewar tree to show only pucks that the user is authorized to see. 
The code needs 2 env variables, "STAFF_GROUP" and "NSLS_API_URL". STAFF_GROUP checks if the user running the code is part of staff. "NSLS_API_URL" point to the NSLS API that exposes proposal information.

Refactored and optimized code by reducing long functions and splitting them, also put repeating code in its own function to be reused. Initial testing works.

### DIFF
--- a/config_params.py
+++ b/config_params.py
@@ -1,4 +1,5 @@
 from enum import Enum
+import os, grp
 
 # BlConfig parameter variable names
 
@@ -109,3 +110,8 @@ VALID_TRANSMISSION = {
 }
 
 LSDC_SERVICE_USERS = ("lsdc-amx", "lsdc-fmx", "lsdc-nyx")
+IS_STAFF = (
+    True
+    if os.environ["STAFF_GROUP"] in [grp.getgrgid(g).gr_name for g in os.getgroups()]
+    else False
+)
--- a/gui/dewar_tree.py
+++ b/gui/dewar_tree.py
@@ -35,7 +35,7 @@ class DewarTree(QtWidgets.QTreeView):
         self.setDragDropMode(QtWidgets.QAbstractItemView.InternalMove)
         self.setAnimated(True)
         self.model = QtGui.QStandardItemModel()
-        
+
         # self.isExpanded = 1
         self.setContextMenuPolicy(QtCore.Qt.CustomContextMenu)
         self.customContextMenuRequested.connect(self.openMenu)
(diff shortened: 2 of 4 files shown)


config_params.py
```diff
@@ -109,3 +110,8 @@ class RasterStatus(Enum):
 }
 
 LSDC_SERVICE_USERS = ("lsdc-amx", "lsdc-fmx", "lsdc-nyx")
+IS_STAFF = (
+    True
+    if os.environ["STAFF_GROUP"] in [grp.getgrgid(g).gr_name for g in os.getgroups()]
```
JunAishima, line 115: STAFF_GROUP should be required

gui/dewar_tree.py
```diff
@@ -89,173 +97,181 @@ def keyPressEvent(self, event):
...
+            current_index = mountedIndex
+
+        if current_index:
+            self.setCurrentIndex(current_index)
+            self.parent.row_clicked(current_index)
+
+    def is_proposal_member(self, proposal) -> bool:
+        # Check if the user running LSDC is part of the sample's proposal
+        if proposal not in self.proposal_membership:
+            r = requests.get(f"{os.environ['NSLS_API_URL']}/proposal/{proposal}")
```
JunAishima, line 233: NSLS_API_URL should be renamed to NSLS2_API_URL and must be required (check it).
also, the "proposal" field name is too generic - this should be called proposal_id

JunAishima (changes requested):
the main body of the work seems good, but the parameters should be required so that the unexpected behavior when they are empty are avoided

gui/dewar_tree.py
```diff
@@ -89,173 +97,181 @@ def keyPressEvent(self, event):
...
+                sample_id,
+                sampleDict.setdefault(sample_id, db_lib.getSampleByID(sample_id)),
+            )
+
+            #if not IS_STAFF and not self.is_proposal_member(sample["proposalID"]):
+            if not self.is_proposal_member(sample["proposalID"]):
+                # If the user is not part of the proposal or is not staff, don't fill tree
+                # Clear the puck information and don't make it selectable
+                # parentItem.clearData()
+                parentItem.setText(index_label)
```
JunAishima, line 170: a question - should this be made more visibly occupied but not available to the current user? it wouldn't surprise me if, for example, beamline staff like to know what the arrangement of pucks is, instead of it being effectively being shown as empty

JunAishima (approved):

PR #339, opened 2023-10-18
Fixed displaying puck name with proposal number

vshekar:
Added to master since faster-dewar-refresh is not in 2023-3-amxfmx

### DIFF
--- a/gui/dewar_tree.py
+++ b/gui/dewar_tree.py
@@ -181,7 +181,8 @@ class DewarTree(QtWidgets.QTreeView):
                 )
                 return
 
-            parentItem.setText(f"{index_label} pass-{sample['proposalID']}")
+            if not parentItem.text().endswith(f"pass-{sample['proposalID']}"):
+                parentItem.setText(f"{parentItem.text()} pass-{sample['proposalID']}")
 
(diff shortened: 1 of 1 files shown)


JunAishima (commented):
I don't know whether I like this setting of the parentItem() by extending the string - seems like it may be prone to creating multiple strings added on, if there is a situation where, for whatever reason, this setting code is called after the puck name is already set.

do you see how situations where this extra adding-on might happen? if it is possible, can you think of a way to mitigate this issue?

vshekar:
The need for adding the pass number arises in such a way is because proposal numbers are not associated with puck names, so we can only get the pass number in the loop that cycles through sample information. 

Puck name is already set outside of this loop, the only thing left to do is add the pass information. The `if parentItem.text().endswith()`  statement checks if the pass info is already added to the string, if it is, nothing more to do. I I guess instead of endswith() check I can replace it with `if j = 0` to only add the string in the first iteration of the for loop, if you think that is more appropriate.

JunAishima:
looks like refreshTreeDewarView() is self-contained enough that you are reconstructing the name from scratch each time and should be safe enough.

JunAishima (approved):

PR #340, opened 2023-10-19
Added queue-collect to user screen

vshekar:


JunAishima (commented):
I would like some more information about some of the decisions that have been made here.

- Is there a reason why the queue collect should be on both the user and staff screens?
- Why is the gripper unmount while cold ability only available for staff? 
- Can you explain how the showing of each of the queue collect check boxes is going to be controlled? I see "show" for both, but only staff seems to be usable at the moment.
- parts of `queueCollectOnCheckBoxCB()` and `checkQueueCollect()` seem to be redundant - wouldn't the setting of queue collect checkbox in `checkQueueCollect()` result in `queueCollectOnCheckCB()` to be called? if so, isn't the changing of queue collect status redundant?

vshekar:
- I've added it in both screens mostly because of muscle memory for staff. Or people who are not aware of when this move will come (they might forget that it is moved to user screen and I will expect a panicked slack message asking where it is)
- Gripper unmount cold is also available in the user screen here https://github.com/NSLS-II/lsdc/blob/b5f4fcf90eb91303d0c576a3651561e42e4c34bf/gui/dialog/user_screen.py#L56
- Both the staff screen and user screen are initialized once at startup and are shown/hidden when you click the appropriate button. Therefore if the queue collect is activated in user screen, the staff screen doesn't update by default (staff screen only checks the queue collect status at init). Therefore the show() method is overridden to first check the value of queue collect in the database and then update the staff screen. Same logic the other way round, user screen will check values before showing the window.
- Yes setting of queuecollect check box in `checkQueueCollect()` will trigger `queueCollectOnCheckCB()` I can remove the status

JunAishima:
I'm concerned about implementations diverging when duplicated. Can you get more input on whether you should be able to implement features like this without having to keep old versions around? The duplication needs to be removed sometime and training should reduce the shock of the change.

JunAishima (approved):

PR #346, opened 2023-11-13
Replaced python calls with json

vshekar:
Added a mechanism to replace exec and function calls with a structured json that can only execute functions in a whitelist

### DIFF
--- a/daq_main_common.py
+++ b/daq_main_common.py
@@ -1,186 +1,264 @@
+import _thread
+import atexit
+import json
+import logging
+import os
 import string
 import sys
-import os
 import time
-import _thread
+import traceback
+
+import beamline_lib
+import beamline_support
+import config_params
 import daq_lib
 import daq_utils
 import det_lib
-import beamline_support
-import beamline_lib
-from start_bs import plt
-import config_params
-import atexit
-import traceback
-
-#imports to get useful things into namespace for server
-from daq_macros import *
-from daq_lib import *
-from robot_lib import *
 from beamline_lib import *
-from gov_lib import setGovRobot
+from daq_lib import *
 
-import logging
+# imports to get useful things into namespace for server
+from daq_macros import *
+from gov_lib import setGovRobot
+from robot_lib import *
+from start_bs import gov_robot, plt
+
 logger = logging.getLogger(__name__)
 
 sitefilename = ""
-global command_list,immediate_command_list,z
+global command_list, immediate_command_list, z
 command_list = []
 immediate_command_list = []
 z = 25
 
-  
+
+def setGovState(state):
+    setGovRobot(gov_robot, state)
+
+
+functions: "List[Callable]" = [
+    anneal,
+    set_beamsize,
+    importSpreadsheet,
+    mvaDescriptor,
+    setTrans,
+    loop_center_xrec,
+    autoRasterLoop,
+    snakeRaster,
+    ispybLib.insertRasterResult,
+    backlightBrighter,
+    backlightDimmer,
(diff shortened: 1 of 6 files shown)


vshekar:
Will force push without the code formatting, as there are too many changes to review

JunAishima (approved):
Overall, looks good. There are also a few other positive aspects:

- A lot of args/kwargs preserve numerical type, if the value isn't coming from a GUI text box
- Get rid of a fair bit of ugly old-style string assembly, before f strings
- generate_server_message() is well-composed to handle command, args, kwargs to allow full flexibility of calls as with the previous method
- no ispybLib commands are on the whitelist, except for the insertRasterResult() so it shouldn't be possible to do the full cycle of adding yourself to a different proposal. you probably could add your collection to another experiment, but that won't be too useful (but would need to reverse-engineer all of the information in the collection request to do so)
- Do you think we could just extract the send_to_server() function so that one function will be called to generate the message, log, then send to the appropriate queue?

gui/control_main.py
```diff
@@ -5202,24 +5165,41 @@ def controlEnabled(self):
             and self.controlMasterCheckBox.isChecked()
         )
 
-    def send_to_server(self, s):
-        if s == "lockControl":
+    def send_to_server(self, function_name: str, args: "Optional[List]" = None, kwargs: "Optional[Dict]" = None):
```
JunAishima, line 5168: Can you explain why you incorporated this function as part of the class? There isn't anything inherently specific to the class, so I would have left it as a utility function.

JunAishima (commented):
Just want some more explanation for why the generate_server_message() needed to be brought into the ControlMain class

gui/control_main.py
JunAishima: realized that the generate_server_message() would only be called within send_to_server() now, so would only be required as part of ControlMain(), and confirmed by @vshekar ... I think it makes sense as it is.

daq_main_common.py
```diff
@@ -29,10 +31,64 @@
...
 def execute_command(command_s):
-  logger.info('executing command: %s' % command_s)
-  exec(command_s)
+  logger.info("executing command: %s" % command_s)
+  try:
+      command: "dict[str, Any]" = json.loads(command_s)
+      func = whitelisted_functions[command["function"]]
+      func(*command["args"], **command["kwargs"])
+  except Exception as e:
+      logger.error(f"Error executing {command_s}: {e}")
```
danielballan, line 91: Would it be help to use `logger.exception` here, which shows both the log message and the full traceback? I suppose it depends on how user-facing these logs are.

daq_main_common.py
```diff
@@ -29,10 +31,64 @@
...
+
+whitelisted_functions: "Dict[str, Callable]" = {
+    func.__name__: func for func in functions
+}
   
 def execute_command(command_s):
-  logger.info('executing command: %s' % command_s)
-  exec(command_s)
+  logger.info("executing command: %s" % command_s)
+  try:
```
danielballan, line 86: Consider making separate try/except blocks for _parsing_ the command (deserialize and function look-up) and _executing_ it (`func(...)`). The message `Error executing ....` could be misleading here.

danielballan (approved):

daq_main_common.py
vshekar: Addressed in [79071fe](https://github.com/NSLS-II/lsdc/pull/346/commits/79071fec2b4b13dd2640ddd57b3ac34f52bcaae7)

danielballan (approved):

PR #347, opened 2023-11-27
Making NYX dewar 8x3. Also changes naming conventions when opening dewar gui

NYSBC-Rudra:
Later fixes:
- Make naming on dewar gui consistent after adding a puck
- Make dewar 6x5 and change the calculations that use /3 (or %3) to get dewar dimensions


JunAishima:
this is a huge number of changes - could you explain a bit how you got to this point? you may need to break this up into a couple of separate pull requests, one where you merge the main master into md2, then your changes... or something along those lines. the main issue I have is that it's hard to see what is a current change and what is a previous change - so combining all of the updates to "our" current master could be one pull request, then your changes another. let me know if you need help with this.

PR #348, opened 2023-11-28
Prevent prefilling of parameters when request clicked

vshekar:
- Hovering over request shows parameters
- Right click menu has option to use params

### DIFF
--- a/gui/control_main.py
+++ b/gui/control_main.py
@@ -4500,9 +4500,6 @@ class ControlMain(QtWidgets.QMainWindow):
         energy_s = str(daq_utils.wave2energy(reqObj["wavelength"], digits=6))
         dist_s = str(reqObj["detDist"])
         self.detDistMotorEntry.getEntry().setText(str(dist_s))
-        # self.dataPathGB.setFilePrefix_ledit(str(reqObj["file_prefix"]))
-        # self.dataPathGB.setBasePath_ledit(str(reqObj["basePath"]))
-        # self.dataPathGB.setDataPath_ledit(str(reqObj["directory"]))
 
     def refreshCollectionParams(self, selectedSampleRequest, validate_hdf5=True):
         reqObj = selectedSampleRequest["request_obj"]
--- a/gui/control_main.py
+++ b/gui/control_main.py
@@ -395,6 +395,7 @@ class ControlMain(QtWidgets.QMainWindow):
         colStartLabel.setFixedWidth(140)
         colStartLabel.setAlignment(QtCore.Qt.AlignCenter)
         self.osc_start_ledit = QtWidgets.QLineEdit()
+        self.setGuiValues({"osc_start": "0.0"})
         self.osc_start_ledit.setFixedWidth(60)
         self.osc_start_ledit.setValidator(QtGui.QDoubleValidator())
         self.colEndLabel = QtWidgets.QLabel("Oscillation Range:")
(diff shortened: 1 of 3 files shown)


gui/control_main.py
```diff
@@ -4498,9 +4498,12 @@ def refreshCollectionParams(self, selectedSampleRequest, validate_hdf5=True):
         energy_s = str(daq_utils.wave2energy(reqObj["wavelength"], digits=6))
         dist_s = str(reqObj["detDist"])
         self.detDistMotorEntry.getEntry().setText(str(dist_s))
-        self.dataPathGB.setFilePrefix_ledit(str(reqObj["file_prefix"]))
-        self.dataPathGB.setBasePath_ledit(str(reqObj["basePath"]))
-        self.dataPathGB.setDataPath_ledit(str(reqObj["directory"]))
+        # self.dataPathGB.setFilePrefix_ledit(str(reqObj["file_prefix"]))
+        # self.dataPathGB.setBasePath_ledit(str(reqObj["basePath"]))
+        # self.dataPathGB.setDataPath_ledit(str(reqObj["directory"]))
```
JunAishima, line 4505: If these are no longer required, please remove entirely.

gui/data_loc_info.py
```diff
@@ -45,6 +45,7 @@ def __init__(self, parent: "ControlMain"):
         self.hBoxDPathParams2.addWidget(self.prefix_ledit)
         self.dataNumstartLabel = QtWidgets.QLabel("File Number Start:")
         self.file_numstart_ledit = QtWidgets.QLineEdit()
+        self.file_numstart_ledit.setText("1")
```
JunAishima, line 48: Let's create a Jira ticket and/or pull request to talk with beamline scientists to remove file numbers entirely.

JunAishima (commented):
Along with the specific issues below, can you go over the workflow and what this pull request solves? I would be happy with a demonstration at the beamline if that is useful.

vshekar:
> Along with the specific issues below, can you go over the workflow and what this pull request solves? I would be happy with a demonstration at the beamline if that is useful.

It solves the problem of replacing experiment parameters with the parameters of the collection that is selected from the dewar view. Apparently its not needed by the beamlines. But it provides an option to view the parameter on hovering over the request. If the user wants to use those parameters its available as a right click option.

gui/control_main.py
vshekar: Addressed in https://github.com/NSLS-II/lsdc/pull/348/commits/0d74496d71ff9d936f82847f08a850e6228d28e8

gui/control_main.py
```diff
@@ -395,6 +395,7 @@ def createSampleTab(self):
         colStartLabel.setFixedWidth(140)
         colStartLabel.setAlignment(QtCore.Qt.AlignCenter)
         self.osc_start_ledit = QtWidgets.QLineEdit()
+        self.setGuiValues({"osc_start": "0.0"})
```
JunAishima, line 398: were the fields of oscillation start, file number start, and resolution not shown when the request is not pre-filled?

vshekar: Correct, previously everytime the dewar tree refreshes, it would find out the current mounted sample and current request, then trigger a row_clicked programatically which would prefill the parameters. This would happen at GUI start too. Now since that doesn't happen on startup those values need a default

PR #354, opened 2024-01-15
Return empty image if eiger monitor is empty

vshekar:


JunAishima (commented):
can you explain what happened if no image was provided to Albula? would it just crash?

gui/albula/controller.py
```diff
@@ -156,14 +156,18 @@ def set_current_monitor_image(self):
...
+                self.ip, self.api_version
+            )
+            replyData = requests.get(urlData)
+            img_bytes = BytesIO(replyData.content)
+            img = Image.open(img_bytes)
+            array = numpy.array(img)
+            array[array == 65535] = 0
+        except:
+            logger.error("Could not get image from EIGER")
+            array = numpy.zeros((3110, 3269))
```
JunAishima, line 170: this looks like one of the 9Ms (Eiger 9M and Eiger2 9M are slightly different sizes).

could you get the actual size from AreaDetector? that would ensure that FMX and NYX get the correct size image back (assuming that this is for Eiger 9M, on AMX)

JunAishima (changes requested):

gui/albula/controller.py
JunAishima: doesn't seem like the actual size of the image is too important, so resolving as a won't-fix

PR #357, opened 2024-01-20
Added default text for raster step

vshekar:
Added a default value for raster step since if you add any other collection at startup, LSDC will complain that not all parameters are filled even though the added collection has nothing to do with rasters

### DIFF
--- a/gui/control_main.py
+++ b/gui/control_main.py
@@ -723,6 +723,7 @@ class ControlMain(QtWidgets.QMainWindow):
         self.rasterStepEdit = QtWidgets.QLineEdit(str(self.rasterStepDefs["Coarse"]))
         self.rasterStepEdit.textChanged[str].connect(self.rasterStepChanged)
         self.rasterStepEdit.setFixedWidth(60)
+        self.rasterStepEdit.setText("20")
         self.rasterGrainRadioGroup = QtWidgets.QButtonGroup()
         self.rasterGrainCoarseRadio = QtWidgets.QRadioButton("Coarse")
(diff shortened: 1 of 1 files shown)


JunAishima:
It would be better to fix the validation/error issue - can you show the exception here?

vshekar:
> It would be better to fix the validation/error issue - can you show the exception here?

The exception happens here: https://github.com/NSLS-II/lsdc/blob/cfdd2a0fc61e54a593b398673dc5164f8fdffd21/gui/control_main.py#L4073

This is within the function that creates a request dictionary. The function assumes that all text boxes are filled otherwise will throw an exception.
One issue is that it is inside a generic try block so its difficult to say from the error message where the exception happened. I used a debugger to pin point it. The current work around to this issue that the scientists are using is when the GUI is started, don't add a collection to queue, but instead switch to raster parameters so that the text box gets populated

The self.beamWidth_ledit and self.beamHeight_ledit text boxes were most likely part of the GUI, but replaced with rasterStepEdit. Unfortunately the textboxes were not properly removed, hidden from the GUI, instead used as variables within the code. So when rasterStepEdit is changed it triggers `rasterStepChanged`. This function simply sets beamWidth_ledit and beamHeight_ledit which is then used to create the request in the snippet above.
https://github.com/NSLS-II/lsdc/blob/cfdd2a0fc61e54a593b398673dc5164f8fdffd21/gui/control_main.py#L2290

I think removing these text boxes will be a more involved PR as its being used in multiple places

JunAishima (approved):

PR #359, opened 2024-02-20
Moving omega is done by server and only in SA

vshekar:
Users tend to move omega before the sample finishes alignment. This moves the governor into M state.

This PR, removes the poking of the omega directly from the GUI. The server moves omega using bluesky and only in SA state


JunAishima (approved):
nice to move a command over to the service side, which decreases direct communication via EPICS. also good to see you use a plan, though in the future, this should go onto a real queue instead of running a separate RE instance

PR #361, opened 2024-03-05
Fixed proposal membership query so that production API server is used

vshekar:


gui/dewar_tree.py
```diff
@@ -272,19 +272,19 @@ def add_samples_to_puck_tree(
             self.parent.row_clicked(current_index)
 
     def is_proposal_member(self, proposal_id) -> bool:
-        # Check if the user running LSDC is part of the sample's proposal
-        if proposal_id not in self.proposal_membership:
-            r = requests.get(f"{os.environ['NSLS2_API_URL']}/proposal/{proposal_id}")
+        # Check if the user running LSDC is part of the sample's proposal 
+        # If the proposal_membership set is empty get data from API (Use prod API)
+        # This way API is only polled once when GUI starts
```
JunAishima, line 277: so if the proposal info gets updated while the user is running the program, they would need to shut down. this should only be a problem if they weren't on the visit originally. possibly an issue if they are removed during the current visit.

we could maybe force the user to reload info in situation 1, but what about situation 2 (no incentive to reload)? a timed retrieval of the info?

gui/dewar_tree.py
```diff
@@ -272,19 +272,19 @@ def add_samples_to_puck_tree(
...
-            ]:
-                self.proposal_membership[proposal_id] = True
-            else:
-                logger.info(f"Users not found in response: {response}")
-                self.proposal_membership[proposal_id] = False
-        return self.proposal_membership[proposal_id]
+            for session in response.get("data_sessions", []):
+                # Assuming data_sessions has values of the form "pass-123456"
+                proposal = session.split("-")[1]
+                self.proposal_membership.add(proposal)
```
JunAishima, line 285: actually, this won't take care of case 2 above, since you only add proposals, never remove. so would require clearing all proposals and re-creating the set.

gui/dewar_tree.py
vshekar: Few reasons why I didn't address situation 2:
- I am assuming the process of removing someone from PASS is a long one which propagates to the server API with a delay. Has there been a case where a user was removed from a proposal _during_ a visit?
-  The GUI closes automatically when the proposal is changed
- I want to avoid pinging the API server every few seconds (minutes?) or whatever we decide is the interval to keep checking all the users that have the GUI open

JunAishima: the case of somebody being removed from a proposal should be handled - while it won't be common, it's possible this could happen, logically.

any querying of the API server should be on the minutes scale - we don't need to boot anybody immediately, and there is already a lag period of data coming from the PASS API into the NSLS2 API.

vshekar: So what do you suggest the interval should be?

JunAishima (changes requested):

PR #362, opened 2024-03-05
Updated cryostream signal for AMX

vshekar:
Set up cryostream temp value to the GUI

### DIFF
--- a/config_params.py
+++ b/config_params.py
@@ -83,7 +83,7 @@ GOVERNOR_TIMEOUT = 120  # seconds for a governor move
 DEWAR_SECTORS = {"amx": 8, "fmx": 8, "nyx": 5}
 PUCKS_PER_DEWAR_SECTOR = {"amx": 3, "fmx": 3, "nyx": 3}
 
-cryostreamTempPV = {"amx": "AMX:cs700:gasT-I", "fmx": "FMX:cs700:gasT-I"}
+cryostreamTempPV = {"amx": "XF:17IDB-ES:AMX{CS:1}SAMPLE_TEMP_RBV", "fmx": "FMX:cs700:gasT-I"}
 
 VALID_EXP_TIMES = {
     "amx": {"min": 0.005, "max": 1, "digits": 3},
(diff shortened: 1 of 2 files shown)


gui/control_main.py
```diff
@@ -142,7 +142,7 @@ class ControlMain(QtWidgets.QMainWindow):
     roiChangeSignal = QtCore.Signal(int, str)
     highMagCursorChangeSignal = QtCore.Signal(int, str)
     lowMagCursorChangeSignal = QtCore.Signal(int, str)
-    cryostreamTempSignal = QtCore.Signal(str)
+    cryostreamTempSignal = QtCore.Signal(object)
```
JunAishima, line 145: is there any reason why it couldn't be a float or something?

config_params.py
```diff
@@ -83,7 +83,7 @@ class RasterStatus(Enum):
 DEWAR_SECTORS = {"amx": 8, "fmx": 8, "nyx": 5}
 PUCKS_PER_DEWAR_SECTOR = {"amx": 3, "fmx": 3, "nyx": 3}
 
-cryostreamTempPV = {"amx": "AMX:cs700:gasT-I", "fmx": "FMX:cs700:gasT-I"}
+cryostreamTempPV = {"amx": "XF:17IDB-ES:AMX{CS:1}SAMPLE_TEMP_RBV", "fmx": "FMX:cs700:gasT-I"}
```
JunAishima, line 86: FMX's PB does not appear to be working right now (March 8, 10:23am) - can this be updated now?

JunAishima (changes requested):

config_params.py
JunAishima: any update on the actual PV that works for FMX?

vshekar: No, FMX does not have a working system, they are going to install it during the upcoming shutdown in which case I will update this PV

gui/control_main.py
vshekar: For FMX currently the value returned by the PV is None whereas its a float at AMX. This is a catch all for different values. It is correctly converted to string in the GUI. Once the FMX PV provides correct values we can switch it to float

config_params.py
JunAishima: good enough for now then...

JunAishima (approved):

JunAishima:
created this issue for future reference, about the cryostream PV: https://github.com/NSLS-II/lsdc/issues/374

PR #368, opened 2024-04-04
3 click centering monitor now starts with waiting for three click

NYSBC-Rudra:
Thought I was just adding line 1441 but hopefully this will still work

### DIFF
--- a/gui/control_main.py
+++ b/gui/control_main.py
@@ -139,7 +139,7 @@ class ControlMain(QtWidgets.QMainWindow):
     fastShutterSignal = QtCore.Signal(float)
     gripTempSignal = QtCore.Signal(float)
     ringCurrentSignal = QtCore.Signal(float)
-    threeClickSignal = QtCore.Signal(float)
+    threeClickSignal = QtCore.Signal(str)
     sampleExposedSignal = QtCore.Signal(float)
     sampMoveSignal = QtCore.Signal(int, str)
     roiChangeSignal = QtCore.Signal(int, str)
--- a/gui/control_main.py
+++ b/gui/control_main.py
@@ -139,7 +139,7 @@ class ControlMain(QtWidgets.QMainWindow):
     fastShutterSignal = QtCore.Signal(float)
     gripTempSignal = QtCore.Signal(float)
     ringCurrentSignal = QtCore.Signal(float)
-    beamAvailableSignal = QtCore.Signal(float)
+    threeClickSignal = QtCore.Signal(float)
     sampleExposedSignal = QtCore.Signal(float)
     sampMoveSignal = QtCore.Signal(int, str)
     roiChangeSignal = QtCore.Signal(int, str)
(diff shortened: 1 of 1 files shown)


gui/control_main.py
```diff
@@ -3058,6 +3086,8 @@ def measurePolyCB(self):
     def center3LoopCB(self):
         logger.info("3-click center loop")
         self.threeClickCount = 1
+        self.threeClickSignal.emit('{} more clicks'.format(str(4-self.threeClickCount)))
```
JunAishima, line 3089: I would suggest using f strings see section 7.1.1 - https://docs.python.org/3/tutorial/inputoutput.html . the text to emit would look like '{4-self.threeClickCount} more clicks'

gui/control_main.py
```diff
@@ -2220,13 +2243,18 @@ def processRingCurrent(self, ringCurrentVal):
         else:
             self.ringCurrentMessage.setStyleSheet("background-color: #99FF66;")
 
-    def processBeamAvailable(self, beamAvailVal):
-        if int(beamAvailVal) == 1:
+    '''
```
JunAishima, line 2246: beam available is supposed to be a global signal of whether the sample would be exposed to X-rays if the shutter were open, regardless of what the user is doing on the GUI.

gui/control_main.py
```diff
@@ -3820,9 +3875,19 @@ def pixelSelect(self, event):
                 correctedC2C_x = correctedC2C_x * scale_x
                 correctedC2C_y = correctedC2C_y * scale_y
                 self.md2.centring_click.put(f"{correctedC2C_x} {correctedC2C_y}")
+                #logger.info('waiting for motor rotation')
```
JunAishima, line 3878: remove commented-out text - if necessary, use git history to retrieve it, unless it is likely to be used soon

gui/control_main.py
```diff
@@ -139,7 +139,7 @@ class ControlMain(QtWidgets.QMainWindow):
     fastShutterSignal = QtCore.Signal(float)
     gripTempSignal = QtCore.Signal(float)
     ringCurrentSignal = QtCore.Signal(float)
-    beamAvailableSignal = QtCore.Signal(float)
+    threeClickSignal = QtCore.Signal(str)
```
JunAishima, line 142: since it looks like the beam available GUI feature is being replaced with three-click, I suggest using a completely new GUI object instead of re-using an old one. Re-using items with the same name is going to be difficult to debug in the feature than if it is correctly named.

gui/control_main.py
```diff
@@ -3832,9 +3897,35 @@ def pixelSelect(self, event):
...
+            self.threeClickLines = []
+
         return
 
+    '''
+    Function to check if MD motors are rotating or not
+    '''
+    def omegaMoveCheck(self, sleeptime,call='OmegaState'):
+        state = self.md2.exporter.read(call)
+        while(state == 'Moving'):
```
JunAishima, line 3916: if these kinds of checks are going to be done in multiple areas of the code, I suggest moving the function into a more central place and have the states as enums or constants (like things in config_params.py) so that small issues such as wrong comparison of strings ("string" != "String") don't lead to functional issues.

PR #370, opened 2024-04-06
Remove 3 click centering for AMX and FMX

vshekar:


JunAishima (approved):
we should check with NYX whether they want this - if not, we may want to permanently remove this.

PR #373, opened 2024-04-09
Set UNMOUNT_COLD_CHECK to True when server starts

vshekar:
Both AMX and FMX have requested that unmount cold at the end of auto collections be available by default.

### DIFF
--- a/gui/dialog/staff_screen.py
+++ b/gui/dialog/staff_screen.py
@@ -337,7 +337,6 @@ class StaffScreenDialog(QtWidgets.QFrame):
         else:
             setBlConfig("queueCollect", 0)
             self.gripperUnmountColdCheckBox.setEnabled(False)
-            self.gripperUnmountColdCheckBox.setChecked(False)
             self.parent.queue_collect_status_widget.setText("Queue Collect: OFF")
         self.parent.row_clicked(
             0


gui/dialog/staff_screen.py
```diff
@@ -337,7 +337,6 @@ def queueCollectOnCheckCB(self, state):
         else:
             setBlConfig("queueCollect", 0)
             self.gripperUnmountColdCheckBox.setEnabled(False)
```
JunAishima, line 339: this is the kind of case where using a signal that can be followed by the GUI makes sense. is it worth it to change types of signal?

vshekar: I'm not sure if I understand

JunAishima: I mean that this kind of signal, because there is no link between the server setting and the GUI value, can get out of sync. if, for example, one of the Comm PVs is used, that could be monitored and kept synchronized between server and GUI

vshekar: Would this be a separate PV for each server setting? How would this work?
Also, you're right maybe this should be a separate issue discussion. My personal opinion is that we should move to a websocket architecture asap as it would eliminate most of the sync issues

JunAishima: yeah, the PV method isn't really scalable in the long-term. please open an issue (or put somewhere within plans) for using websockets. resolving as done

JunAishima (approved):

PR #382, opened 2024-05-23
Server Heartbeat

mskinner5278:
This is a utility for notifying the user of server failures. After 10 seconds of the client not finding a server, it will send a popup through the popup pv.

### DIFF
--- a/daq_main_common.py
+++ b/daq_main_common.py
@@ -43,7 +43,8 @@ def pybass_init():
   daq_lib.init_var_channels()
   if getBlConfig(config_params.DETECTOR_OBJECT_TYPE) != config_params.DETECTOR_OBJECT_TYPE_NO_INIT:
     det_lib.init_detector()  
-  daq_lib.message_string_pv = beamline_support.pvCreate(daq_utils.beamlineComm + "message_string")    
+  daq_lib.message_string_pv = beamline_support.pvCreate(daq_utils.beamlineComm + "message_string")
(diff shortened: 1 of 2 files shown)


JunAishima:
you have reminded me that I was going to try to improve the LSDC server service so that it will try to restart in case it dies, and is able to restart if there isn't a chronic issue. I'll handle that in a separate issue - will have to be through ansible

daq_main_common.py
```diff
@@ -93,11 +95,17 @@ def process_command_file(command_file_name):
 
 
 def process_immediate_commands(frequency):
+  memory_beat = beamline_support.pvGet(daq_lib.heartbeat)
   while (1):
     if (len(immediate_command_list) > 0):
       command = immediate_command_list.pop(0)
       logger.info('immediate command: %s' % command)
       process_input(command)
+    if memory_beat == beamline_support.pvGet(daq_lib.heartbeat):
```
JunAishima, line 104: is there any reason you couldn't have this in a separate thread or something? it seems unnecessarily tied to a particular function, and while this is unlikely, the checking interval for the `process_immediate_functions()` could be set in such a way that the popup would always trigger.

JunAishima (changes requested):

daq_main_common.py
mskinner5278: I only did it to minimize the creation of new threads, but if you think that is a non-issue then I can split it out.

PR #387, opened 2024-06-06
Changing 4 beamsize options to 2

vshekar:
V0H0 -> S
V1H1 -> L
V0H1 and V1H0 removed

### DIFF
--- a/daq_macros.py
+++ b/daq_macros.py
@@ -39,7 +39,7 @@ import bluesky.plan_stubs as bps
 import bluesky.plans as bp
 from bluesky.preprocessors import finalize_wrapper, finalize_decorator
 from fmx_annealer import govStatusGet, govStateSet, fmxAnnealer, amxAnnealer # for using annealer specific to FMX and AMX
-from config_params import ON_MOUNT_OPTION, OnMountAvailOptions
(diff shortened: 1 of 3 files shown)


JunAishima (changes requested):
could you extract the beam_mapping to somewhere like config_params, then use the keys and values of that dictionary for the comparisons in set_beamsize() and beamSizeOptionList? that would keep all of the configuration info in one place, reducing chance of unsynchronized settings

daq_macros.py
```diff
@@ -4077,12 +4077,9 @@ def set_beamsize(sizeV, sizeH):
     setPvDesc("CRL_H1B_IN",1)
   else:
     logger.error("Error: Horizontal size argument has to be \'H0\' or  \'H1\'")
-  if (sizeV == 'V0' and sizeH == 'H0'):
-    daq_lib.set_field("size_mode",0)
-  elif (sizeV == 'V1' and sizeH == 'H1'):
-    daq_lib.set_field("size_mode",1)
-  else:
-    pass
+  for index, key in list(BEAMSIZE_OPTIONS.keys()):
```
JunAishima, line 4080: do you mean `enumerate` instead of `list`?

JunAishima (changes requested):

JunAishima (approved):

PR #390, opened 2024-06-24
Vector drag

vshekar:
Added feature to drag and drop vector nodes on the sample camera

Previous workflow to add vector was:
- "Click to center" to the position where vector starts
- Add Vector start node
- "Click to center" where vector ends
- Add Vector end node
- Rotate sample by 90 degrees
- "Click to center" for each node separately and adjust vector start or end

New workflow:
- "Add quick vector" button will place a complete vector of specified size
- User drags vector start and end nodes without moving motors
- Rotate sample by 90 degrees
- User adjusts vector start and end nodes by dragging without moving motors

Or:
- "Add start node" places node at center, user drags node to desired position
- "Add end node" places end node at center, also dragged to adjust
- Same process orthogonally

Previous workflow still works for users who don't know

All vector related functions and data are moved to gui/vector.py which manages internal state of the vector


JunAishima (approved):
nice consolidation of code into the vector module!

PR #393, opened 2024-07-01
Robot recovery procedure during automated collection for EMBL robot

vshekar:
There have been multiple cases where during automated collection, the robot starts the unmounting procedure. Just before the robot gripper picks up the sample, it waits for the governor to get into SE state. If it does not reach SE (a common case being the piny or pinz not moving, additionally the GovMon not recovering from the error in time). 

In this case the gripper stays at the unmount position indefinitely. This PR detects the SE timeout error and runs a robot recovery macro, which includes:
1. Recovering robot
2. Drying gripper
3. Park and cool gripper
After the recovery procedure, it will attempt to unmount again (by this time GovMon should have done its magic) and continue the automated collection. 
If the second attempt fails as well, the recovery macro is run and the user is informed and automated collection is stopped

### DIFF
--- a/embl_robot.py
+++ b/embl_robot.py
@@ -459,6 +459,7 @@ class EMBLRobot:
                 RobotControlLib.unmount2(absPos)
               except Exception as e:
                 # If there is an exception again, return UNMOUNT_FAILURE
+                daq_macros.run_robot_recovery_procedure()
                 e_s = str(e)
                 message = "ROBOT unmount2 ERROR: " + e_s
                 daq_lib.gui_message(message)
--- a/embl_robot.py
+++ b/embl_robot.py
@@ -451,7 +451,7 @@ class EMBLRobot:
               daq_macros.disableMount()
               daq_lib.gui_message(e_s + ". FATAL ROBOT ERROR - CALL STAFF! robotOff() executed.")
               return UNMOUNT_FAILURE
-            elif (e_s.find("SE") != -1):
+            elif "SE" in e_s:
               # In case there is an SE Timeout error, run the recovery procedure
               daq_macros.run_robot_recovery_procedure()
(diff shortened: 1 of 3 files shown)


embl_robot.py
```diff
@@ -450,10 +451,25 @@ def unmount(self, gov_robot, puckPos, pinPos, sampID):
               daq_macros.disableMount()
               daq_lib.gui_message(e_s + ". FATAL ROBOT ERROR - CALL STAFF! robotOff() executed.")
               return UNMOUNT_FAILURE
-            message = "ROBOT unmount2 ERROR: " + e_s
-            daq_lib.gui_message(message)
-            logger.error(message)
-            return UNMOUNT_FAILURE
+            elif (e_s.find("SE") != -1):
```
JunAishima, line 454:
```suggestion
            elif "SE" in e_s:
```

embl_robot.py
```diff
@@ -450,10 +451,25 @@ def unmount(self, gov_robot, puckPos, pinPos, sampID):
...
+              # In case there is an SE Timeout error, run the recovery procedure
+              daq_macros.run_robot_recovery_procedure()
+              try:
+                # Try to unmount again
+                RobotControlLib.unmount2(absPos)
+              except Exception as e:
+                # If there is an exception again, return UNMOUNT_FAILURE
+                daq_macros.run_robot_recovery_procedure()
+                e_s = str(e)
+                message = "ROBOT unmount2 ERROR: " + e_s
```
JunAishima, line 464:
```suggestion
                message = f"ROBOT unmount2 ERROR: {e}"
```

JunAishima (changes requested):

JunAishima (approved):

PR #396, opened 2024-07-16
Added mounting status indicator with Enum

vshekar:
This PR adds a status indicator to the sample list. Will display whether the sample is currently being mounted, finished mounting, currently unmounting etc.

This status helps beamline scientists recover easily when things go wrong and they can tell at what point things went wrong


JunAishima (approved):
was useful when I was testing other code on LSDC. having the status attached to the sample is useful because this highlights which sample is being handled at the moment

PR #397, opened 2024-07-17
Data security guard rails

vshekar:
Fixing multiple edge cases where data can potentially go in the wrong directory. These issues can only be caused if staff don't follow prescribed steps and accidentally run into these edge cases.

Edge case 1:
When "Queue collect" is on it allows staff to add requests to multiple samples. These samples __could__ belong to different proposals. We only want to add requests to samples of the currently active proposal. The commit will not allow adding requests for proposals that exist in the API, if it doesn't exist in the API (for e.g. 999999) it will ask the user if they are sure about adding the request. Code is in commit: https://github.com/NSLS-II/lsdc/commit/e02dba893b71c38b892d33b7a78c2c2f5a43e5c0

Edge case 2:
When staff schedules an auto raster to a sample belonging to proposal A while proposal A is active but accidentally collect after switching to proposal B. The two rasters will go to directory B but the standard will go to directory A. This is because any new request created uses the currently active proposal directory. The commit will fix it so that rasters will use the path of the standard collection instead of blindly using the current proposal directory. Code is in commit:  c9ca3097c2159c7ff896dd9592544849bbb9a935

Paths used are resolved so that paths with /data/ can be compared to /data4/. Commit in 67ba3533d8c4bb2cbf47491c29f003972af6ae67 and c0a61703320a7433b3f163a1e62394d6ef4f69aa

Edge case 3:
Because there can be a large number of pucks and samples that can be loaded at the same time, staff can sometimes forget that they have added requests to a proposal that is NOT the current proposal. i.e. Add the request to a sample for proposal A when its active but forget to collect or remove it after switching to proposal B. When the user pushes collect queue in this situation, popNextRequest will check if the queue has samples from multiple proposals. If it finds that there are requests from multiple proposals AND the commissioning directory is not used, it will return a None, otherwise will return the next collection to run. Commit in cba5c95d66be202f188f1df39640fd4d3eeb530f

Related fixes:
Dewar tree will show which directory the data will go to in the pop up box for collections (cba5c95d66be202f188f1df39640fd4d3eeb530f)
Prevent LSDC from crashing if a proposal number doesn't exist in NSLS2 API (cba5c95d66be202f188f1df39640fd4d3eeb530f)
Message to let user know why the queue didn't run (4e23d203b969f05b5de8fcca48b252d2ed0182f1)


JunAishima:
Edge case 1 - looks good, makes sense
Edge case 2 - looks good
thanks for handling the data path resolution - a new surprise!
Edge case 3 - looks like it works - logging fixes suggested

db_lib.py
```diff
@@ -663,13 +663,29 @@ def popNextRequest(beamlineName):
...
+    request_paths = set()
+    for req in orderedRequests:
+        if req["priority"] > 0 and req["priority"] != 99999:
+            request_proposal_ids.add(req["proposalID"])
+            request_paths.add(req["request_obj"]["basePath"])
+    visit_dir_path = Path(getBeamlineConfigParam(beamlineName, "visitDirectory")).resolve()
+    if request_proposal_ids:
+        if len(request_proposal_ids) > 1 and "commissioning" not in visit_dir_path.parts:
+            # Multiple proposal requests being run and not in commissioning
+            logger.info(f"Visit dir parents: {visit_dir_path.parts}")
```
JunAishima, line 677: this and the following info statements should not display path info to current users.

db_lib.py
```diff
@@ -663,13 +663,29 @@ def popNextRequest(beamlineName):
...
+            return None
```
JunAishima, line 678: the default return type appears to be {}, as seen at the end of the function. any reason why this is different? (also applies for the following elif)

gui/control_main.py
```diff
@@ -4072,25 +4104,47 @@ def addRequestsToAllSelectedCB(self):
...
-            self.EScanDataPathGB.setFilePrefix_ledit(
-                str(self.selectedSampleRequest["request_obj"]["file_prefix"])
-            )
-            self.EScanDataPathGB.setDataPath_ledit(
-                str(self.selectedSampleRequest["request_obj"]["directory"])
-            )
-            self.addSampleRequestCB(selectedSampleID=self.selectedSampleID)
+            sample_data = db_lib.getSampleByID(self.selectedSampleID)
+            prop_dir = self.get_proposal_directory(sample_data["proposalID"])
+            add_request = True
```
JunAishima, line 4109: could you extract the `add_request` calculation to outside of this code without breaking up the flow when reading the code?

JunAishima (changes requested):

gui/control_main.py
vshekar: Moved to separate method in 330a9f9d014e950048771cf79166f8df996ba26d

JunAishima (approved):

PR #401, opened 2024-12-17
Display fastdp summary table in GUI

vshekar:
Add 2 widgets to display fastdp results
- Text widget that monitors a text file on disk and directly displays it (LogViewerWidget)
- Table widget that monitors a csv file and displays it as a qttablewidget (CSVTableWidget)

Current implementation of summarytable program only updates the text file live. Therefore LSDC uses the logviewerwidget to show the table. In the future when summarytable updates the csv live we will switch to the CSVTableWidget. 

The LogViewerWidget is useful for any file on disk (such as log files)


JunAishima (changes requested):
Just one small change to make sure you don't get unexpected types to try to display.

I especially like the auto-update feature.

JunAishima (approved):

PR #422, opened 2026-03-23
Remove kafka

JunAishima:
Kafka is to be decommissioned at NSLS-II soon. This PR removes all code associated with sending messages to Kafka topics.


vshekar (approved):
I think these are the only references to kafka, looks good to me

